PR #187, opened 2014-02-26
initial jasmine2 support (mostly working for the html reporter).

jejacks0n:
jasmine2 doesn't want to play nice.. and this is hard. sadly.


gylaz:
Any progress on this? This would be sweet!

PR #252, opened 2014-07-26
Add support for configuring path to istanbul executable

jcarlson:
In my Rails project, istanbul is packaged as a Node dependency. The path to the executable is then `./node_modules/.bin/istanbul`.

Some devs on the team have this path in their shell `PATH` var, but some do not. As there are some breaking API changes with the latest istanbul (0.3.0), it would be nice to tell Teaspoon to use the version bundled with the app and not bother searching the system path.

I wouldn't imagine this would be terribly difficult, would it?

### DIFF
--- a/spec/teaspoon/command_line_spec.rb
+++ b/spec/teaspoon/command_line_spec.rb
@@ -112,6 +112,7 @@ describe Teaspoon::CommandLine do
         **** Coverage ****
 
           -C, --coverage CONFIG_NAME       Generate coverage reports using a pre-defined coverage configuration.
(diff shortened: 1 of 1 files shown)


derekpeterson:
I was going to work on adding support for this, and found this PR while looking for discussion, so I'm also curious about this feature's status.

I would like to have this option to avoid relying on global installs. Another hack is to prepend `node_modules/.bin` to the `PATH`, but that's not really a great solution. Official support would be better.

(Anyway, thanks for a great tool!)

mikepack:
We don't have any immediate plans to work on this, but would gladly accept a PR.

PR #263, opened 2014-09-09
Prevent early exit after an error

aasmith:
Use the provided timeout to give the runtime longer before quitting after an error.

This fixes a problem where we were seeing a non-fatal parse error prevent our test suite from running.

The error sets a one second timer, during which time, if there are no calls to `console.log`, then a fatal error is assumed. Making this timeout tunable gives us some leeway in troubleshooting.


jejacks0n:
since I couldn't easily merge this, and it was a simple change I've made the change and will see where net out. thanks.

PR #277, opened 2014-11-02
Switching instrument_file? to be in Instrumentation class

golmansax:
This was done to solve the RequireJS issue (https://github.com/modeset/teaspoon/pull/240), would like to hear your thoughts about this.  Some things:

* The reason why this change works with RequireJS is because the decision to instrument is now made in the Sprockets hook, and all RequireJS requests go through Sprockets.
* I wasn't sure how to grab the current config so I grabbed the default config with `Teaspoon.configuration.suite_configs['default'][:instance]`.  I can help write specs once this part is figured out.
* I was thinking about completely forgoing the addition of `instrument=1`, but wasn't sure which way to go.  Either way works with RequireJS, as you can add a `instrument=1` to the end of every RequireJS request through the `urlArgs` config.

### DIFF
--- a/app/views/teaspoon/suite/_boot_require_js.html.erb
+++ b/app/views/teaspoon/suite/_boot_require_js.html.erb
@@ -3,11 +3,9 @@ rails_config = Rails.application.config
 require_options = {baseUrl: rails_config.assets.prefix}
 require_options.merge!(rails_config.requirejs.user_config) if rails_config.respond_to?(:requirejs)
 specs = @suite.spec_assets(false).map{ |s| "#{s.gsub(/\.js.*$/, "")}" }
-
-specs.each {|s| require_options['shim'][s] = [@suite.helper] }
 %>
 
-<%= javascript_include_tag 'require' %>
+<%= javascript_include_tag @suite.helper %>
 <script type="text/javascript">
   Teaspoon.onWindowLoad(function () {
     // setup the Teaspoon path prefix to load /assets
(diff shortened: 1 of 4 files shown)


jejacks0n:
I'll give this a try and revert if needed.

jejacks0n:
I removed the stuff you had shifted out of suite, because it was only using the default configuration -- which broke some specs, and seems short sighted.

jejacks0n:
I now understand what the changes were, but don't fully understand the need for them.. can you explain further?

jejacks0n:
If I understand correctly, you need to instrument files that are being loaded via requirejs, and thus the instrument=1 param isn't present. I'll poke around a bit, and understand the desire for it, but absolutely abhor having to do shims in teaspoon to support requirejs -- it's at an entirely different level, and one I happen to disagree with.

If it can come at a different layer, it would be better. So the ideas I've kicked around, but can't quite resolve are to add an additional configuration -- but even that doesn't work well, and the core reason is that the instrumentation determination has to come at the level when a file is requested -- with all the context of the configuration used during the test run -- which is not possible when an asset is requested.

davestevens:
I've just opened #315 that simply adds an option to RequireJS from within the boot view which adds `instrument=1` to all requests made through require.

golmansax:
I'm sorry I probably can't explain the reasoning as well as I could have, since I haven't looked at this code for a while now haha.  I think your analysis is pretty on point, here is my best guess to what I was thinking at the time:

- I forget exactly why, but I believe the requirejs files didn't get the `&instrument=1`  through `Teaspoon::Suite#asset_url`.  However, you can around it in requirejs by adding `instrument=1` to all files through the config like what @davestevens said
- So then, all of the files have `instrument=1`, but the `no_coverage` files were not filtered out.  I guess this is why I switched the `no_coverage` check to `Teaspoon::Instrumentation` (which can filter the requirejs files because requirejs also goes through Sprockets)

Sorry for the code sloppiness - this was meant to be more of exploratory pull than one with production ready code.  Just wanted to get your thoughts on it because I wasn't really considering how this would impact the non-requirejs use cases of teaspoon.

PR #305, opened 2015-02-09
Update Mocha link on README

TrangPham:
The old link is broken and leads to a 404 page. 

Also it's not clear from the README if it's Mocha JS being used or Mocha gem: https://github.com/freerange/mocha is being used. This confused me at first because I haven't had exposure to Mocha JS before, and thus did not presume that's what was meant.


jejacks0n:
Thanks!

Issue #301, opened 2015-01-28
Default location of requiring js files is not reachable if you uncomment chai assertion libraries in spec_helper.js


kareemamin:
In the default spec_helper.js file if you decide to use Chai and uncomment the following variable assignment:

    // window.assert = chai.assert;
       window.expect = chai.expect;
    // window.should = chai.should();

Then the following section which loads all js in the rails app seems to be unreachable i.e. the files that I expect to be loaded that are in application.js are not loaded. 

    // You can require your own javascript files here. By default this will include everything in application,      however you
    // may get better load performance if you require the specific files that are being used in the spec that tests them.
    //= require application

However, if I move the require application directive to the top of the file it loads the correct files. Not sure if the preprocessor stops once it sees js code but if I comment the variable assignment for Chai it loads the file as expected.

I think it's worthwhile to move the loading of other js files to the top along with the require directives for support libraries to avoid this problem.

jejacks0n:
You're probably right. Do you want to submit a PR?

---
Jeremy Jackson

kareemamin:
Sounds good I'll do that

jejacks0n:
Thanks for helping sort out the confusion!

PR #310, opened 2015-02-22
move the default position of the require comment to the top.

igwan:
I just fell into the catch described by kareemamin here https://github.com/modeset/teaspoon/issues/301.
I looked for error elsewhere in my code for way to long, not realizing the require position was the problem. So here is a PR.


jejacks0n:
Thanks, but please adjust this to only shift the application require to not be above any dependency loads.. you've put it above the bind-poly in some cases, and it should not be above any other dependency.

Issue #258, opened 2014-08-18
Document TEASPOON_ENV variable


crismali:
I figure this should be mentioned in the comments of the `teaspoon_env.rb` templates and the readme. I would have submitted a pull request but I'm not sure exactly where this should go in the readme (I'm still happy to submit a pr)

mikepack:
Documented in #314

PR #314, opened 2015-02-23
Document TEASPOON_ENV

mikepack:
Addresses #258

Hoping this covers all cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,7 +374,7 @@ end
 
 ## Configuration
 
-When you install Teaspoon a `teaspoon_env.rb` file is generated that contains most of this information, but we've provided it here too.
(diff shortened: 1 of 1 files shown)


jejacks0n:
Seems reasonable. I don't disagree with where you have it, but is it maybe better suited for the CLI area of the documentation? I guess the reason is that if the env is not where it's expected, it can be manually required within the rails pipeline so the browser interface still works, but the env is one of the first things loaded from the CLI, and so that's when the override comes into play.

mikepack:
This override wouldn't affect the browser runner? I would suspect putting this in a `.env` would have it affect all uses of Teaspoon.

jejacks0n:
That's true it would. Point taken.

PR #316, opened 2015-02-23
Ensure QUnit tests have a message

mikepack:
I don't know that we need the complexity of #233, but we should at least produce a message in the case that it's not present.

### DIFF
--- a/app/assets/javascripts/teaspoon/qunit.coffee
+++ b/app/assets/javascripts/teaspoon/qunit.coffee
@@ -33,6 +33,7 @@ class Teaspoon.Spec
     return [] unless @spec.failed
     for item in @spec.assertions
       continue if item.result
+      @provideFallbackMessage(item)
       {message: item.message, stack: item.source}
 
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/teaspoon/qunit.coffee
```diff
@@ -33,6 +33,7 @@ class Teaspoon.Spec
     return [] unless @spec.failed
     for item in @spec.assertions
       continue if item.result
+      item.message ||= "Expected #{JSON.stringify(item.actual)} to equal #{JSON.stringify(item.expected)}"
```
jejacks0n: I haven't configured coffeescript/javascript stuff on hound yet, so feel free to ignore.

mikepack:
Worth noting: QUnit produces decent error messages in other cases, just not with `#equal`.

app/assets/javascripts/teaspoon/qunit.coffee
jejacks0n: is there an easy way to test this? these tests live in test/javascript

mikepack:
@jejacks0n tests added

chancancode:
:heart_eyes: @mikepack Thank you for doing this!

mikepack:
@chancancode not a problem, thanks for all your work on this. We'll be releasing 0.9 soon, and this change will go out with that release.

Issue #308, opened 2015-02-12
Version 1.0 Roadmap


mikepack:
- [x] Pull testing libraries out of core. Install would be `gem install teaspoon-mocha`. (breaking)
- [x] Support Jasmine 2.0
- [x] Support Mocha 2.0
- [x] Namespace framework-specific implementations instead of monkey patching (breaking)
- [x] Register formatters with core
- [x] Register drivers with core
- [x] Rework coverage configuration (breaking)

PR #322, opened 2015-03-01
[WIP] Ignored instrumentation files

davestevens:
This is an initial attempt at including a configuration option to filter files being instrumented by Istanbul.

I've added the configuration at the top level as the `coverage` configuration is based on a suite which it not available from within `Teaspoon::Instrumentation`.
This is currently allowing me to define a list of strings which are then used as regular expressions to check against the assets pathname.

Initial discussion was here: #315.

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,13 +360,11 @@ end
 
 Teaspoon allows defining files and directories to ignore when performing coverage instrumentation.
 
-The below example will ensure that the files within `/test/javascripts` and `/vendor/assets` are not intrumented by Istanbul and will not show up in the coverage results.
+The below example will ensure that any files within `/test/javascripts` and `/vendor/assets` are not intrumented by Istanbul and will not show up in the coverage results.
 
 ```ruby
-config.coverage_ignored = ["/test/javascripts", "/vendor/assets"]
+config.coverage_ignored = ["/test/javascripts", %r{/vendor/assets}]
 ```
-
-The defined strings are used within regular expression and any filepath which matches any of these will not be instrumented.
 
 ## Configuration
 
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ Teaspoon allows defining files and directories to ignore when performing coverag
 The below example will ensure that the files within `/test/javascripts` and `/vendor/assets` are not intrumented by Istanbul and will not show up in the coverage results.
 
 ```ruby
-config.coverage_ignored_files = ["/test/javascripts", "/vendor/assets"]
+config.coverage_ignored = ["/test/javascripts", "/vendor/assets"]
 ```
 
 The defined strings are used within regular expression and any filepath which matches any of these will not be instrumented.
--- a/README.md
+++ b/README.md
@@ -356,6 +356,17 @@ config.coverage :CI do |coverage|
 end
 ```
 
+### Ignoring Files
+
+Teaspoon allows defining files and directories to ignore when performing coverage instrumentation.
+
+The below example will ensure that the files within `/test/javascripts` and `/vendor/assets` are not intrumented by Istanbul and will not show up in the coverage results.
+
+```ruby
+config.coverage_ignored_files = ["/test/javascripts", "/vendor/assets"]
+```
+
+The defined strings are used within regular expression and any filepath which matches any of these will not be instrumented.
 
 ## Configuration
 
--- a/lib/teaspoon/configuration.rb
+++ b/lib/teaspoon/configuration.rb
@@ -26,21 +26,21 @@ module Teaspoon
 
     cattr_accessor :driver, :driver_options, :driver_timeout, :server, :server_port, :server_timeout, :fail_fast,
                    :formatters, :color, :suppress_log,
-                   :use_coverage, :coverage_ignored_files
-    @@driver                 = "phantomjs"
-    @@driver_options         = nil
-    @@driver_timeout         = 180
-    @@server                 = nil
-    @@server_port            = nil
-    @@server_timeout         = 20
-    @@fail_fast              = true
+                   :use_coverage, :coverage_ignored
+    @@driver           = "phantomjs"
+    @@driver_options   = nil
+    @@driver_timeout   = 180
+    @@server           = nil
+    @@server_port      = nil
+    @@server_timeout   = 20
+    @@fail_fast        = true
 
-    @@formatters             = ["dot"]
-    @@color                  = true
-    @@suppress_log           = false
+    @@formatters       = ["dot"]
+    @@color            = true
+    @@suppress_log     = false
 
-    @@use_coverage           = nil
-    @@coverage_ignored_files = []
+    @@use_coverage     = nil
+    @@coverage_ignored = []
 
     # options that can be specified in the ENV
 
@@ -153,10 +153,10 @@ module Teaspoon
       @@formatters.to_s.split(/,\s?/)
     end
 
-    def self.coverage_ignored_files
-      return [] if @@coverage_ignored_files.blank?
-      return @@coverage_ignored_files if @@coverage_ignored_files.is_a?(Array)
-      @@coverage_ignored_files.to_s.split(/,\s?/)
+    def self.coverage_ignored
+      return [] if @@coverage_ignored.blank?
+      return @@coverage_ignored if @@coverage_ignored.is_a?(Array)
+      @@coverage_ignored.to_s.split(/,\s?/)
     end
 
     # override from env or options
--- a/lib/teaspoon/configuration.rb
+++ b/lib/teaspoon/configuration.rb
@@ -153,12 +153,6 @@ module Teaspoon
       @@formatters.to_s.split(/,\s?/)
     end
 
-    def self.coverage_ignored
-      return [] if @@coverage_ignored.blank?
-      return @@coverage_ignored if @@coverage_ignored.is_a?(Array)
-      @@coverage_ignored.to_s.split(/,\s?/)
-    end
-
     # override from env or options
 
     def self.override_from_options(options)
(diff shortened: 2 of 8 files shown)


jejacks0n:
Looks like a good start.. can we put this in the coverage specific configuration? Seems like it's pretty much directly tied to that.

jejacks0n:
I appreciate the effort, and will merge it when it's moved into that area of the configuration. =)

davestevens:
I was unsure about where it would go in the coverage specific configurations.
If it goes within the `Coverage` class then its linked to a suite, which is unknown from within `Instrumentation`. Where do you think it should go? If it put it in `Coverage` and then look up the configuration assuming `default` as the suite name?

jejacks0n:
haha, fair enough. I hadn't thought about that. pfff... thinking on that, but thanks for making the point.

jejacks0n:
@mikepack how do you feel about this conversation? Do you think it's really bad if we set the current_suite at a Teaspoon level from within the controller so the instrumentation can determine what files to instrument or not? We're hobbled by how sprockets hooks work.

README.md
```diff
@@ -356,6 +356,17 @@ config.coverage :CI do |coverage|
...
 ```
 
+### Ignoring Files
+
+Teaspoon allows defining files and directories to ignore when performing coverage instrumentation.
+
+The below example will ensure that the files within `/test/javascripts` and `/vendor/assets` are not intrumented by Istanbul and will not show up in the coverage results.
+
+```ruby
+config.coverage_ignored_files = ["/test/javascripts", "/vendor/assets"]
```
mikepack: The configuration option here is `coverage_ignored_files` but these are directories. I wonder if this would be better as just `coverage_ignored`.

For the sake of consistency, we should probably make this configuration more like the `matcher` config:

```ruby
suite.matcher = "{spec/javascripts,app/assets}/**/*_spec.{js,js.coffee,coffee}"
```

lib/teaspoon/instrumentation.rb
```diff
@@ -14,7 +14,14 @@ def self.add?(response, env)
         env["QUERY_STRING"].to_s =~ /instrument=(1|true)/ &&            # the instrument param was provided
         response[0] == 200 &&                                           # the status is 200 (304 maybe?)
         response[1]["Content-Type"].to_s == "application/javascript" && # the format is something that we care about
-        response[2].respond_to?(:source)                                # it looks like an asset
+        response[2].respond_to?(:source) &&                             # it looks like an asset
+        !ignored?(response[2])                                          # it should not be ignored
+    end
+
+    def self.ignored?(asset)
+      Teaspoon::Configuration.coverage_ignored_files.select do |file|
```
mikepack: This should use `#any?` instead of `#select` so that the looping will short circuit if it finds one that matches.

lib/teaspoon/configuration.rb
```diff
@@ -152,6 +153,12 @@ def self.formatters
       @@formatters.to_s.split(/,\s?/)
     end
 
+    def self.coverage_ignored_files
+      return [] if @@coverage_ignored_files.blank?
```
mikepack: Is it necessary for this line to be here? Since the default is an empty array, we should be safe, no?

spec/teaspoon/configuration_spec.rb
```diff
@@ -149,6 +152,19 @@
 
   end
 
+  describe ".coverage_ignored_files" do
+
+    it "returns an empty list if nothing was set" do
+      expect(subject.coverage_ignored_files).to eq([])
+    end
+
+    it "returns an array of  if they were comma separated" do
```
mikepack: @jejacks0n do we do similar things elsewhere? Do we need to support arrays and comma-delimited strings?

jejacks0n: Yes, but only on the CLI/ENV overrides.
https://github.com/modeset/teaspoon/blob/578fd4cd089aede2983e628f1abe60cfcb1a602a/lib/teaspoon/configuration.rb#L147

mikepack:
@jejacks0n I'm very unfamiliar with this part of the codebase. We should stay away from a `current_suite`, IMO. What prevents us from calling `Teaspoon.coverage_configs`? Is it that we need to know what coverage config is being used (eg `default` or `CI`)?

spec/teaspoon/configuration_spec.rb
jejacks0n: Here's another case, but this doesn't feel like a configuration for the CLI, or one that should have an ENV:
https://github.com/modeset/teaspoon/blob/578fd4cd089aede2983e628f1abe60cfcb1a602a/lib/teaspoon/configuration.rb#L138-138

mikepack: Right, because we can't pass an array in in those cases. Do CLI flags get passed in as configuration? ie does `teaspoon --coverage_ignored_files="/spec/javascripts,/vendor/assets"` and this will automagically work?

jejacks0n: sadly no, because they kinda need to be defined directly -- so you can use the short version (-c etc).

mikepack: If this configuration can't be used on the CLI, I don't think we should support the string version.

mikepack: In fact, I still think we should support a style similar to the `matcher` config.

jejacks0n:
Because you can specify a coverage config at runtime, or by configuring one within your suite.

README.md
jejacks0n: Agree here, though those are slightly different. We have an asset, which should respond to logical_path, which gives you the full pathname of the file that's going to be served. I'm unsure if you can do a thing to see if the filepath we have would match that sort of dir glob string.

spec/teaspoon/configuration_spec.rb
mikepack: If we're going to support two forms, I think we should support regexes.

README.md
mikepack: Ah, got it. Yeah, maybe that style won't work as well here.

spec/teaspoon/configuration_spec.rb
jejacks0n: agreed, which is all I support with the suite.no_coverage configuration.

jejacks0n: ah, I take that back -- I support an array of strings, or regex.

mikepack: So, curious, why do we need this if there is a `suite.no_coverage` option? Didn't know that existed.

jejacks0n: here's where that behavior is exhibited: https://github.com/modeset/teaspoon/blob/1ed8cfe34a4abe6375df619f1bb595b39749a2b7/lib/teaspoon/suite.rb#L41-41

jejacks0n: ok, so when a suite is rendered via the controller, we usually tell it which files to include, and at that time we can determine if we want to instrument it or not (by adding an "instrument=true" param to the asset). Then when the asset is loaded the instrumenter checks for that param and does or does not instrument the file (the instrumenter hooks into sprockets).

With requirejs, when those assets are requested, requirejs doesn't know which ones should have the param, and the instrumenter doesn't know either because it doesn't know what suite was rendered.

jejacks0n:
I'm thinking setting the current_suite is really the only way to accomplish it.. so, I dislike requirejs, have I said that already? haha. ;-P

lib/teaspoon/instrumentation.rb
davestevens: Ah, good catch.

lib/teaspoon/configuration.rb
davestevens: This was a case of copying the way which `formatters` work. You're right, if it were blank it would be an empty array anyway.

davestevens:
@jejacks0n thats fine with me. I'll make changes based on the way the matcher works and the other feedback.

README.md
davestevens: I agree about the option name, I was unsure while I was writing it.
The matcher idea should work here, the dir glob should be able to match the pathname of the assets. I'll have a go and see.

spec/teaspoon/configuration_spec.rb
davestevens: This was another case of seeing the way `formatters` work (as its an array on the same level) and using that. I think the matcher style is better and should work.

README.md
davestevens: I was just thinking about the matcher and I think it makes more sense here to use a similar style to the `no_coverage` option. Then you would be able to ignore specific directories or files. I don't think with the matcher it could ignore both `spec/javascripts` and `vendor/assets`?

jejacks0n:
Ok, so I appreciate your effort on this, and sorry I haven't merged it yet.

There seems to be a solution that'll work better than this, and you have my apology for leading you down a less optimal path. I'm happy to do this work myself now, but I'll explain it anyway in case you want to give it a shot -- it's not far from this implementation really.

Ok, so instead of putting it at a top level configuraiton, it should be a coverage configuration option.. essentially just moving no_config from suite to coverage. The reason for this, is that the coverage configuration is known globally, and so after some thought it seems like it should technically belong within the coverage configuration anyway. This change does involve more changes than what's in this PR to make the suite understand these concepts through the coverage configuration.

davestevens:
Thats fine, and it does make sense to have this option on the coverage configuration.

This is where I had originally put the option, but I ran into the issue of being unaware of which suite was currently being run when `Teaspoon:: Instrumentation.add?` was executed. What is your idea for this?

jejacks0n:
It's not associated to the suite. It's set at the global use_coverage configuration level. That configuration can be specified at run time as well.

---
Jeremy Jackson

mikepack:
@davestevens thanks to this PR, we've had some discussions internally and have plans to rework the `coverage` configuration for our 1.0 release (#308). We're thinking of removing all coverage options at the suite level, ie. moving the `no_coverage` to the coverage block.

Going to close this PR. Happy to reopen if you want to take a stab at getting this into the `coverage` block. Otherwise, we'll address it in the not-so-distant future.

PR #326, opened 2015-03-05
Require javascripts to be defined in suite configs.

jejacks0n:
This forces javascripts to be specified.. currently not very useful, but will become more useful during the transition phase where users may not have a framework installed/required.


mikepack:
:+1:

PR #333, opened 2015-03-15
Provides a structure for frameworks to share test harnesses.

jejacks0n:
This shifts integration specs, and javascript specs into their respective gems.

- [x] further cleanup of specs
- [x] get build passing
- [x] move framework javascripts into their respective gems, and have the tests run on the coffeescripts
- [ ] provide build tools for the frameworks


Rakefile
```diff
@@ -11,17 +11,15 @@ end
 # -----------------------------------------------------------------------------
 APP_RAKEFILE = File.expand_path("../spec/dummy/Rakefile", __FILE__)
 load "rails/tasks/engine.rake"
-Bundler::GemHelper.install_tasks
+begin
+  Bundler::GemHelper.install_tasks
+rescue RuntimeError
```
mikepack: What's this fallback for?

jejacks0n: so, we share this Rakefile for development -- should it be desired. And it will contain the basic build task. So, if it's being loaded from teaspoon-qunit, it does the right thing, but if it's being loaded from the root (aka Teaspoon), it doesn't know which gemspec to use for the release commands since there's two (teaspoon.gemspec and teaspoon-dev.gemspec)

PR #339, opened 2015-04-01
Add ES6 extensions to normalize_js_extension.

phyllisstein:
This super-insignificant tweak adds `.es6` and `.js.es6` to the preprocessors filtered out by the `normalize_js_extension` method. I'd like to come back to it at some point and see if it's at all feasible to futureproof it by extracting preprocessor extensions from the `suite.matcher` setting, but don't have the time just now.

### DIFF
--- a/lib/teaspoon/suite.rb
+++ b/lib/teaspoon/suite.rb
@@ -97,7 +97,7 @@ module Teaspoon
     end
 
     def normalize_js_extension(filename)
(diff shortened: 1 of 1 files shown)


jejacks0n:
Let's just pull this out and put it in a configuration -- then you can add a wiki article about it if you'd like. I'll include es6 in the default value too. I'll merge this and then refactor to configuration.

Thanks!

PR #340, opened 2015-04-04
Do not double load teaspoon_env

### LINKED ISSUE #332: Teaspoon Hooks are called twice from setup
Selenium (2.45.0 and 2.43.0 tested)
Teaspoon (0.8.0 and 0.9.1 tested)
Browsers tested (firefox, chrome)
Ruby 2.2.0
Rails 4.1.0

I am only calling the hook once on setup of one test and I put a debugger right after that. It still puts the message in console twice.

If needed any other information let me know.

mikepack:
- When loading from the console, we need to load the Rails env, so we load the app's teaspoon_env. When we load teaspoon_env, the app then loads all engines, including teaspoon. Previously, this double loaded teaspoon_env and caused multiplicative effects on configurations that accumulate values, like #hooks.

- Updates the predicate method `rails_loaded?` to always return a boolean

- Converts `environment.rb` to group methods on the singleton (`class << self`)

Fixes #332

### DIFF
--- a/lib/teaspoon/console.rb
+++ b/lib/teaspoon/console.rb
@@ -5,6 +5,7 @@ module Teaspoon
     def initialize(options = {})
       @default_options = options
       @suites = {}
+      Teaspoon::Environment.console = true
       Teaspoon::Environment.load(options)
 
       @server = start_server
--- a/lib/teaspoon/engine.rb
+++ b/lib/teaspoon/engine.rb
@@ -15,7 +15,9 @@ module Teaspoon
 
     initializer :assets, group: :all do |app|
       begin
-        Teaspoon::Environment.require_environment
+        if Teaspoon::Environment.require_environment_from_engine?
+          Teaspoon::Environment.require_environment
+        end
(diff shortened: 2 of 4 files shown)


jejacks0n:
:+1: The only thing I'm so-so on is the `class << self` -- reason being it's more difficult in terms of readability to someone not super experienced in ruby. Up to you on if you want to change it though.

Nice work!

mikepack:
Word. I'll change that and merge.

jejacks0n:
@mikepack I honestly don't mind it, that was just feedback.

mikepack:
It sounds like you really hate it, so I'm going to remove it and never do it in the future. :trollface:

jejacks0n:
haha

mikepack:
FYI, couldn't completely remove `class << self` as it's the only want to define an `attr_accessor` on a module.

jejacks0n:
Fair enough. Somewhat related, we should move away from active_support as we go too. Feel free to do that in other PRs.

PR #343, opened 2015-04-06
Load Rails, which will load teaspoon_env

### LINKED ISSUE #341: You can no longer provide a root configuration and append asset paths when running via the CLI.
If you have an engine for instance it's common to setup a dummy rails app within the spec path. It's nice to not have to put the javascript specs within the dummy app, and instead configure teaspoon to specify it's root, and any additional asset paths that you might want to include.

Things are working with the rake task, but not when run via the CLI. This has to do with the changes to how things are loaded when run via the console -- #340.

mikepack:
If we always load Rails, we can assume that `teaspoon_env.rb` will get loaded as part of the engine initialization. This doesn't help us move away from Rails, but it's the simplest way to ensure that `teaspoon_env.rb` is only being required once, without limiting any behavior.

Additional:
- :+1: Removes the needs for `teaspoon_env.rb` to load the Rails environment which I think is cleaner. If we want to go the direction of this PR, I propose we rename this file to `teaspoon.rb`.
- :-1: We now need to know how to boot Rails. I've added the `TEASPOON_RAILS_ENV` env var to make this overridable. Works with a stock setup, but is required every time we run teaspoon tests using the dummy app's `environment.rb`.
- :-1: Ties us even closer to Rails.
- :+1: Is less code than my previous approach #340 (which I've since reverted in master).

Fixes #341

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -10,6 +10,7 @@ end
 # Dummy App
 # -----------------------------------------------------------------------------
 APP_RAKEFILE = File.expand_path("../spec/dummy/Rakefile", __FILE__)
+ENV["TEASPOON_RAILS_ENV"] = File.expand_path("../spec/dummy/config/environment.rb", __FILE__)
 load "rails/tasks/engine.rake"
 begin
   Bundler::GemHelper.install_tasks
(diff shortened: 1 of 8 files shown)


jejacks0n:
I'm not opposed to this, but let me screw around with it to see how easy it would be to take rails out as a dependency for a bit. It kinda plays into supporting middleman, and potentially providing a simple rack middleware layer that teaspoon uses if rails isn't present -- though that sounds particularly unpleasant. heh.

Issue #191, opened 2014-03-11
Routes are not being prepended


psharpNumerex:
Upgraded to 8 and it seems that the routes are not being prepended.

I tracked it down to engine.rb:43

return if Teaspoon::Engine.routes.recognize_path('/') rescue nil

outputting the results of Teaspoon::Engine.routes.recognize_path('/') gives me
{:controller=>"teaspoon/suite", :action=>"index"}

I'm using Rails 3.2.17

hlascelles:
We did a cleanup of some engines and saw this error arise (with 0.8.0). It turns out that our [Combustion](https://github.com/pat/combustion) folders needed a NOOP *routes.rb* file to trigger the correct append to take place.

That is to say, adding the following file to _spec/internal/config/routes.rb_ solved the problem:

````
Rails.application.routes.draw do
   # Do nothing, except trigger automatic teaspoon engine mount
end
````

jejacks0n:
That's probably true because we use that "event" as the time to hook in.

---
Jeremy Jackson

mikeantonelli:
@hlascelles do you have an example of using teaspoon and combustion to test a Rails engine?

hlascelles:
Apologies @mikeantonelli I've just seen this! No we don't I'm afraid. Just rather a large and convoluted engine. If / when it is broken down I will post one...

PR #349, opened 2015-04-20
Allow alternative catch-all routes to be specified

ahorner:
We currently prepend Teaspoon’s routes in an effort to prevent catch-all
routes from intercepting traffic intended for the Teaspoon engine.

_However_, recent changes have hard-coded “routing_error” as the only
viable catch-all action, meaning that any custom catch-all routing will
prevent Teaspoon’s routes from being prepended. The solution I’ve opted
for here is to allow users to specify their custom catch-all action, so
that Teaspoon can still properly mount when a path is undefined.

### DIFF
--- a/lib/generators/teaspoon/install/templates/env.rb.tt
+++ b/lib/generators/teaspoon/install/templates/env.rb.tt
@@ -1,5 +1,6 @@
 Teaspoon.configure do |config|
   config.mount_at = "/teaspoon"
+  config.catchall = nil
   config.root = nil
   config.asset_paths = ["<%= framework.install_path %>/javascripts", "<%= framework.install_path %>/javascripts/stylesheets"]
   config.fixture_paths = ["<%= framework.install_path %>/javascripts/fixtures"]
--- a/lib/generators/teaspoon/install/templates/env_comments.rb.tt
+++ b/lib/generators/teaspoon/install/templates/env_comments.rb.tt
@@ -3,6 +3,11 @@ Teaspoon.configure do |config|
   # `http://localhost:3000/jasmine` to run your tests.
   config.mount_at = "/teaspoon"
 
+  # Sets an alternative catch-all action to check when mounting Teaspoon routes. For example, if your application
+  # imposes a catch-all like `get "*path", to: "errors#not_found"`, you should set this to "not_found".
+  # Note: Defaults to "routing_error" if nil.
+  config.catchall = nil
+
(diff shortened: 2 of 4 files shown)


mikepack:
I'd actually like to see if we can determine a routing error without depending on the name of the action. I'll do some poking around and let you know what I come up with.

ahorner:
Sounds good. This was causing an immediate issue on a project, so I went with the band-aid approach.

mikepack:
@ahorner what version of Rails are you on?

ahorner:
4.2.0

mikepack:
@ahorner I'm having a hard time reproducing. Specifically, `app.routes.recognize_path(mount_at)` always throws an exception for me, so there is never a check against the `:action`.

Maybe you can fill in some details. What is special about your usage of Teaspoon or your app's configuration or routes? Can you paste applicable code from your app here?

mikepack:
Aside: #191

ahorner:
For my current project, we have a catch-all route defined at the end of the routes file like so:

`routes.rb`
```ruby
  get "*unmatched_route", unmatched_route: /(?!.*rails\/mailers).*/, to: "errors#not_found"
```

In this case, `recognize_path` returns an action of `not_found` (the action for the catch-all route), and teaspoon just silently loads without attempting to prepend its routes. We've managed to avoid the issue for the time being by just adding `teaspoon` to our unmatched route regex, but it's sort of a weird interaction. 

The approach in this PR is probably overkill, but we should at least touch up the documentation.

mikepack:
So, I have a test app that is running Rails 4.2.1 and Teaspoon 0.9.1. I have this as the last route in my `routes.rb`:

```ruby
get '*path', to: "errors#not_found"
```

When I debug into the suspect line, `app.routes.recognize_path(mount_at)`, it *always* throws an exception.

What could be differing? Are you able to reproduce this in an entirely new app?

ahorner:
I'm not having any luck recreating the issue in an isolated test case, so the issue must be a weird interaction with something else in our stack. I poked around for a while without any success; I'll update this again once I figure out exactly where the issue is coming from. In the meantime, sorry for the wild goose chase.

mikepack:
No problem. I appreciate your time trying to track this down!

mikepack:
Hey @ahorner, I'm going to close this issue for now. Happy to reopen if you find anything on your end.

PR #350, opened 2015-04-21
Speedup javascript_include_tag.

francois2metz:
Appending ?body=1 to all assets make javascript_include_tag very slow.

So only append it to instrumented file and let sprockets to his job for
others files.

In our app, performance improvements with this patch are quite impressive (527 JavaScript files so far)

I instrumented _boot.html.erb:

```ruby
<% puts Benchmark.measure { %>
<%= javascript_include_tag *@suite.spec_assets %>
<% } %>
<script type="text/javascript">
  Teaspoon.onWindowLoad(Teaspoon.execute);
</script>
```

Without the patch: `12.500000   0.500000  13.000000 ( 13.000034)`
With the patch: `3.140000   0.120000   3.260000 (  3.256158)`

### DIFF
--- a/lib/teaspoon/suite.rb
+++ b/lib/teaspoon/suite.rb
@@ -75,8 +75,8 @@ module Teaspoon
     end
 
     def asset_url(asset)
-      params = "?body=1"
-      params << "&instrument=1" if instrument_file?(asset.pathname.to_s)
+      params = ""
+      params << "?body=1&instrument=1" if instrument_file?(asset.pathname.to_s)
       "#{asset.logical_path}#{params}"
     end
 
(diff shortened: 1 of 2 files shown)


mikepack:
If you're running Teaspoon >= 0.9, you can use the `expand_assets` suite configuration to speed things up. The downside is it makes code harder to debug as the stack traces all point to one monolithic file.

https://github.com/modeset/teaspoon/wiki/Teaspoon-Configuration#suite-configuration-directives

francois2metz:
Why not speed up thinks for users that don't want to use `expand_assets`? Any reasons to append `body=1` by default? Sprockets will do for you later.

mikepack:
Because Teaspoon runs in the `development` Rails environment, and we want to provide a consistent experience regardless of whether the host app's `config.assets.debug` is set to `true` or `false`.

PR #354, opened 2015-04-29
Add coverage.ignore option in place of suite.no_coverage

mikepack:
Since RequireJS does not use sprockets to require assets, 4ed7fbd was added to put instrumentation on all assets if coverage is requested. `suite.no_coverage` (removed) was used to exclude files from instrumentation as they are compiled in `suite.spec_assets` and included on the page for testing. Since `suite.no_coverage` was only being referenced when requiring assets from sprockets, it had no affect on people using RequireJS. #322 was proposed to ignore files for the case of RequireJS. This is a more formal stab at the same idea.

### Upgrade notes

Anyone doing:

```ruby
suite.no_coverage += /my_file.js/
```

Should now do:

```ruby
config.coverage do |coverage|
  coverage.ignore += /my_file.js/
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,7 +59,7 @@
 * Provide QUnit tests with a default `equal` message
 * Support cli reusing running Rails server
 * Add support for Capybara Webkit
-* Better support for RequireJS
+* Better support for RequireJS and more consistent coverage configuration (@davestevens)
 * Add RSpec HTML formatter
 
 #### Bug Fixes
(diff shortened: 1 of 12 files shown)


jejacks0n:
Redundant curly braces around a hash parameter.

Looks good aside from that.

We should also consider adding a notice in the deprecation saying that `no_coverage` no longer works, with instructions on how to get the same behavior.

mikepack:
@davestevens Teaspoon 1.0 (unreleased) should fix your issue with RequireJS #322

PR #362, opened 2015-05-06
Keep file filter around when grepping

### LINKED ISSUE #327: Retain file filter when specs are additionally filtered by grep param.
When I click a particular spec I expect only it to run.

Currently, the link has grep but not file, and results in other specs running as well (as seen by their console.log output). Our project has some units with quite similar tests, so grep hits more than one.

Instead, I'd like the url file param to be kept, and the grep param added to it.

It looks like this should be fairly easy to do so I'll try to make a PR.

mikepack:
Fixes #327 

This PR does the following:

- Establishes a mixin pattern.
- Creates `Spec` and `Suite` base classes.
- Utilizes the mixin pattern and base classes to provide URL generation that retains the file filter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ### Unreleased
 
+#### Bug Fixes
+
+* Retain file filter when navigating to specs (#327)
+
+
 ### 1.0.2 (5/5/15)
 
 #### Bug Fixes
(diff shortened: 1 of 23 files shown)


mikepack:
@jejacks0n can I get some :eyes: on this?

PR #368, opened 2015-05-18
Write Istanbul result to stdout when the coverage report errors

derekpeterson:
When Istanbul fails, it helps immensely to have the error that it reports. Having it would have helped me avoid a lot of confusion while trying to get coverage reports set up on Jenkins (Istanbul choked on the name of the Jenkins job). I thought about making this a flaggable option, but it seemed too specific for something general like `--[no-]debug`.


mikepack:
Thanks so much! Sorry for the delay 🍻🙌🎉

PR #372, opened 2015-05-28
If the standard rails config/environment isn’t there...

### LINKED ISSUE #371: Cannot use teaspoon because it's assuming config/environment will always be present.
I have a micro rails application -- with my development application defined in a simple `config.ru` file.

```ruby
require "rubygems"
require "bundler/setup"
require "action_controller/railtie"
require "sprockets/railtie"

Bundler.require

module Micro
  class Application < Rails::Application
    config.session_store :cookie_store, :key => "_app_session"
    config.secret_token = "[token]"
    config.active_support.deprecation = :log
    config.consider_all_requests_local = true
    config.encoding = "utf-8"
    config.eager_load = false

    config.assets.enabled = true
    config.assets.version = "1.0"
    config.assets.debug = true

    config.assets.paths = []
    config.assets.paths << "lib/vendor"
    config.assets.paths << "lib/javascripts"
    config.assets.paths << "lib/stylesheets"
  end
end

class ApplicationController < ActionController::Base
  prepend_view_path Rails.application.root.join('examples')

  def page
    render template: params[:page] || 'index'
  end
end

Rails.application.initialize!
Rails.application.routes.draw do
  # rendering pages
  match '/(:page)' => 'application#page', via: :get
end

run Micro::Application rescue nil
```

I then put a `teaspoon_env.rb` in the spec directory, as per the standard -- and require rails in it:

```ruby
unless defined?(Rails) # Set RAILS_ROOT and load the environment.
  ENV['RAILS_ROOT'] = File.expand_path("..", __FILE__)
  load File.expand_path("../../config.ru", __FILE__)
end

Teaspoon.configure do |config|
  config.mount_at = "/teaspoon"
  config.root = nil
  config.asset_paths = ["spec/javascripts", "spec/javascripts/stylesheets"]
  config.fixture_paths = ["spec/javascripts/fixtures"]

  config.suite do |suite|
    suite.use_framework :mocha
  end
end
```

This simple setup no longer works because internally to teaspoon, it's making assumptions about how rails is setup in pathing and file structures.  This seems like a negative thing to tie teaspoon to, and makes it feel more coupled to rails than desired.

```
/usr/local/opt/rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/teaspoon-1.0.2/lib/teaspoon/environment.rb:56:in `require': cannot load such file -- /Users/jejacks0n/Projects/mathhoops/config/environment (LoadError)
```

jejacks0n:
Assume the teaspoon_env will handle loading rails (and avoiding double configuration). fixes #371

This allows alternate rails setups where there isn't a `config/environment.rb` file in your "rails" application.

Here's a [wiki article](https://github.com/modeset/teaspoon/wiki/Micro-Applications) that describes the process.

### DIFF
--- a/lib/teaspoon/environment.rb
+++ b/lib/teaspoon/environment.rb
@@ -53,7 +53,14 @@ module Teaspoon
 
     def self.load_rails
       rails_env = ENV["TEASPOON_RAILS_ENV"] || File.expand_path("config/environment", Dir.pwd)
-      require rails_env
+
+      # Try to load rails, assume teaspoon_env will do it if the expected
+      # environment isn't found.
+      if File.exists?(rails_env)
+        require rails_env
+      else
+        require_environment
+      end
     end
   end
 end
--- a/spec/teaspoon/environment_spec.rb
+++ b/spec/teaspoon/environment_spec.rb
@@ -24,6 +24,13 @@ describe Teaspoon::Environment do
       described_class.load
     end
 
+    it "falls back to loading the teaspoon environment" do
+      allow(subject).to receive(:load_rails).and_call_original
+      allow(File).to receive(:exists?).and_return(false)
+      expect(subject).to receive(:require_environment)
+      described_class.load
+    end
+
     it "aborts if Rails can't be found" do
       expect(subject).to receive(:rails_loaded?).and_return(false)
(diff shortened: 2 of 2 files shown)


jejacks0n:
@mikepack ping.

PR #381, opened 2015-06-26
Fix try/catch on Jasmine 2 so that it works again

mcmire:
In Jasmine 1, you can toggle whether Jasmine catches exceptions by
setting `Jasmine.CATCH_EXCEPTIONS`.

In Jasmine 2, this option [has moved] to the `jasmine.Env` object, and
instead of setting a property, you call the function `catchExceptions`.

[has moved]: https://github.com/jasmine/jasmine/blob/v2.3.4/src/core/Env.js#L138


jejacks0n:
:+1: Thanks Elliot. =) Ignore the failing specs, istanbul changed something in their output.

jejacks0n:
fwiw, I wish mocha had this capability too.

mcmire:
Yeah, it's pretty useful!

PR #385, opened 2015-06-28
Rails.root vs Dir.pwd

brewster1134:
Potential fix for allowing teaspoon to run via a rails server daemon

### DIFF
--- a/lib/teaspoon/environment.rb
+++ b/lib/teaspoon/environment.rb
@@ -31,7 +31,7 @@ module Teaspoon
       env_files = override ? [override] : standard_environments
 
       env_files.each do |filename|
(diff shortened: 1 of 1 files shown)


jejacks0n:
Hmmm.. I'll have to think about this, but thanks for raising the issue and looking at a fix.

jejacks0n:
Can you rebase master to make sure it's passing? A change to istanbul output broke a few specs.

brewster1134:
so after playing around with this, and better familiarizing myself the teaspoon code, i think this PR can be closed....if i set the TEASPOON_ENV env var before starting the server daemon, i can run the tests successfully... it seems you had already built this ability in :smile:

PR #386, opened 2015-07-02
Removed test_files from gemspec

iainbeeston:
I was just looking through the code installed for teaspoon, and I noticed that the /spec/dummy directory is huge. It turns out that rubygems is packaging up not only /spec/dummy, but also all the tempfiles and logfiles there *at the time of packaging*.

The latest consensus is that test_files shouldn't be packaged with a gem anyway (see https://github.com/bundler/bundler/pull/3207) because they're never used, so rather than filtering out /spec/dummy/log and /spec/dummy/tmp I've prevented rubygems from packing test files at all.


jejacks0n:
I'm cool with that.

jejacks0n:
It'll take a new release to pick that up, and I'm not sure we're ready for that just yet, but with the next release.

PR #389, opened 2015-07-12
Support $.parseHtml for more robust fixtures

### LINKED ISSUE #382: Fixture has <tr> and <td> stripped out when not in <table>
I often return HTML partials, such as a table row, in ajax interactions.

Given a fixture named 'whatever.html' that looks like:
```
<tr>
  <td>Some thing</td>
</tr>
```
Then, when calling `loadFixture('whatever.html')` the `tr` and `td`s are stripped, leaving only `Some thing`.

Is it possible to return the verbatim html?

### LINKED ISSUE #387: Don't use innerHTML for fixtures!
Hi guys.

using `innerHTML` in fixture load/add will cause IE8 and below to completely remove `<script>` tags and any other browser to ignore them (wont run as expected).

You have basically 2 workaround to make it behave correctly:

1- Get rid of innerHTML and use a Cross Browser DOM parser:

``` coffee
onLoadComplete: (content) ->
   fragment = parseFragment(content);
   target.appendChild(fragment);
```

2- A working but risky hack on html:

``` coffee
onLoadComplete: (content) ->
  content = content.replace(/<script/g, '<div data-was-script="1"').replace(/script>/g, 'div>');

  # Use this in place of innerHTML cause it has better support:
  target.insertAdjacentHTML('afterbegin', content);

  traverse target, child ->
    if elem.getAttribute('data-was-script')
            src = child.getAttribute('src');
            runningScript = if src then createScriptWithSrc(src) else createScriptWithContent(child.innerHTML);
            insertAfter(child, runningScript);
            child.parentNode.removeChild(child);
```

mikepack:
Support broken HTML with `$.parseHtml`. Can't really write a good test for this since Teaspoon doesn't ship with jQuery, but here is the test I built against, which fails without jQuery:

```javascript
it "supports invalid HTML", ->
  fixture.set("<td>Row</td>")
  expect(document.getElementById("teaspoon-fixtures").innerHTML).toBe("<td>Row</td>")
```

Fixes #387 
Fixes #382

### DIFF
--- a/app/assets/javascripts/teaspoon/fixture.coffee
+++ b/app/assets/javascripts/teaspoon/fixture.coffee
@@ -63,18 +63,22 @@ class Teaspoon.Fixture
 
   putContent = (content) =>
     cleanup()
-    create()
-    window.fixture.el.innerHTML = content
+    addContent(content)
 
 
   addContent = (content) =>
     create() unless window.fixture.el
-    window.fixture.el.innerHTML += content
+
+    if jQueryAvailable()
+      parsed = $($.parseHTML(content, document, true))
+      window.fixture.el.appendChild(parsed[i]) for i in [0...parsed.length]
+    else
+      window.fixture.el.innerHTML += content
 
 
   create = =>
     window.fixture.el = document.createElement("div")
-    window.fixture.$el = $(window.fixture.el) if typeof(window.$) == 'function'
+    window.fixture.$el = $(window.fixture.el) if jQueryAvailable()
     window.fixture.el.id = "teaspoon-fixtures"
     document.body?.appendChild(window.fixture.el)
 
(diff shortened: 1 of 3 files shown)


mikepack:
@jejacks0n :eyes: :pray:

jejacks0n:
I think jquery is available to our repo through jquery-rails, fwiw. So I think a spec is possible.

jejacks0n:
Code looks good. I'm a bit skeptical that it's a good change though. Like, can I now make an invalid fixture, get a passing spec, and get different results in the real world if I used .insert or .html?

This was my struggle with the issue. I guess we'll find out. :)

mikepack:
Yeah, I'm struggling with that as well. It seems like a bad way of using fixtures, unless I'm misunderstanding the use case. Pulling JS down through a fixture seems poor for performance (because of script download/parse blocking), in the case of #387. Testing just the contents of an invalid element, without caring about the surrounding elements that make it valid, also seems like a bad idea and the tests should be moved to a higher level where it's tested against something valid, in the case of #382.

`¯\_(ツ)_/¯`

PR #391, opened 2015-07-15
Support .js.jsx.coffee file extension in Teaspoon::Suite

plongyear:
We are writing specs for React components. We are using the .js.jsx.coffee file extension so we can use jsx in our specs.

When we try to view these files in our suite by browsing to /teaspoon, the files show up as filename.js.jsx.js. Updating `normalize_js_extension` causes the files to show up correctly when viewing the files at /teaspoon.

Is this the preferred way of supporting new file extensions?

### DIFF
--- a/lib/teaspoon/suite.rb
+++ b/lib/teaspoon/suite.rb
@@ -95,7 +95,7 @@ module Teaspoon
     end
 
     def normalize_js_extension(filename)
-      filename.gsub(".erb", "").gsub(/(\.js\.coffee|\.coffee|\.es6|\.js\.es6)$/, ".js")
+      filename.gsub(".erb", "").gsub(/(\.js\.coffee|\.coffee|\.es6|\.js\.es6|\.js\.jsx\.coffee)$/, ".js")
     end
 
     def glob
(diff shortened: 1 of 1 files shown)


jejacks0n:
@plongyear, thanks for this.

@mikepack, we could move that regex to an undocumented, or lightly documented configuration? It feels dirty to me, but I don't see another way around it. Thoughts, ideas?

mikepack:
@jejacks0n seems reasonable as a config, even with full documentation. Though as a config it should be an array of regexes instead of a single regex with ors. Why's it feel dirty?

jejacks0n:
I think a single regex is fine, but is up to whoever gets around to it, which will probably be you. It just is a bummer because we have to resolve files to sprockets urls.

plongyear:
@jejacks0n @mikepack - I agree, a config file sounds like a great idea.

Since the inline regex is the current way file extensions are being resolved, do you think it makes sense to merge this for now until someone is able to get around to implementing the config approach? It would be great if we could get this in to get our specs to run properly.

plongyear:
I also added a commit that supports both the `.js.jsx` and `.js.jsx.coffee` extensions.

mikepack:
@plongyear I'd say it's probably best to point your Gemfile at your fork until we're able to address this correctly.

plongyear:
@mikepack thanks, will do.

jejacks0n:
@mikepack addressed this, so it's now possible via #418. Thanks for the insight and help figuring out what we could to resolve it!

Issue #401, opened 2015-08-14
Coverage does not show up with expand_assets set to false


andresbravog:
opening a new one referencing #357 as is happening again to me.

mikepack:
This test doesn't suffice? https://github.com/modeset/teaspoon/blob/master/spec/teaspoon/suite_spec.rb#L111-L116

andresbravog:
I guess not.

Placing this in suite:

```Ruby
    def asset_url(asset)
      params = []
      params << "body=1" if config.expand_assets
      puts '>> Loading ' + asset.pathname.to_s.split('myreponame').last
      params << "instrument=1" if instrument_file?(asset.pathname.to_s)
      url = asset.logical_path
      url += "?#{params.join("&")}" if params.any?
      puts '>> |===> with url:  ' + url.split('myreponame').last
      url
    end
```

gives the output

```bash
>> Loading /spec/javascripts/spec_helper.coffee
>> |===> with url:  spec_helper.js?instrument=1
>> Loading /spec/javascripts/backbone/apps/profiles/edit/view/header_spec.coffee
>> |===> with url:  backbone/apps/profiles/edit/view/header_spec.js
>> Loading /spec/javascripts/backbone/apps/profiles/show/view/basic_custom_field_header_spec.coffee
>> |===> with url:  backbone/apps/profiles/show/view/basic_custom_field_header_spec.js
>> Loading /spec/javascripts/backbone/apps/profiles/show/view/contact_email_spec.coffee
>> |===> with url:  backbone/apps/profiles/show/view/contact_email_spec.js
>> Loading /spec/javascripts/backbone/apps/profiles/show/view/contact_link_spec.coffee
>> |===> with url:  backbone/apps/profiles/show/view/contact_link_spec.js
>> Loading /spec/javascripts/backbone/apps/profiles/show/view/contact_phone_spec.coffee
>> |===> with url:  backbone/apps/profiles/show/view/contact_phone_spec.js
>> Loading /spec/javascripts/backbone/apps/profiles/show/view/current_experience_spec.coffee
>> |===> with url:  backbone/apps/profiles/show/view/current_experience_spec.js
>> Loading /spec/javascripts/backbone/apps/profiles/show/view/experience_custom_fields_header_spec.coffee
>> |===> with url:  backbone/apps/profiles/show/view/experience_custom_fields_header_spec.js
>> Loading /spec/javascripts/backbone/apps/profiles/show/view/layout_spec.coffee
[...]

PR #402, opened 2015-08-18
Default coverage ignore regex should not include spec_helper

andresbravog:
Related to #401

### DIFF
--- a/spec/teaspoon/instrumentation_spec.rb
+++ b/spec/teaspoon/instrumentation_spec.rb
@@ -116,6 +116,20 @@ describe Teaspoon::Instrumentation do
 
         expect(subject.add?(response, "QUERY_STRING" => "instrument=true")).to be(false)
       end
+      
+      context 'with default ignored files' do
+        it "does not ignore spec_helper file" do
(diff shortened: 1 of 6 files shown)


spec/teaspoon/instrumentation_spec.rb
```diff
@@ -116,6 +116,20 @@ def ignore(paths)
...
+      
+      context 'with default ignored files' do
+        it "does not ignore spec_helper file" do
+          asset = double(source: source, pathname: "path/to/spec_helper.js")
+          
+          expect(subject.add?(response, "QUERY_STRING" => "instrument=true")).to be(true)
+        end
+        
+        it "does ignore any other helper than spec helper" do
+          asset = double(source: source, pathname: "path/to/other_helper.js")
```
andresbravog: this is used by the response let

mikepack:
Hey Andrés, thanks so much for your efforts clarifying this issue. We've decided that the best default behavior is to exclude `spec_helper` from coverage as it remains today, and recommend that people avoid requiring `application` from within `spec_helper`. We've updated the README to reflect this caveat and provide the desired workaround: https://github.com/modeset/teaspoon#caveats

PR #403, opened 2015-08-26
Fix capybara-webkit driver to use synchronize correctly

ryanong:
apparently you have to raise an error not just return false

### DIFF
--- a/lib/teaspoon/driver/capybara_webkit.rb
+++ b/lib/teaspoon/driver/capybara_webkit.rb
@@ -11,18 +11,22 @@ require "teaspoon/driver/base"
 module Teaspoon
   module Driver
     class CapybaraWebkit < Base
+      class TeaspoonNotFinishedError < StandardError; end
       def initialize(_options = nil)
       end
 
       def run_specs(runner, url)
         session.visit(url)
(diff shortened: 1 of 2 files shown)


mikepack:
Thanks Ryan. This code makes sense and I understand where you're going with it, but I don't understand the problem. What was your expected behavior and what didn't work correctly? There are no new test cases added, so I'm not sure what problem you're solving.

ryanong:
Unfortunately the way this is tested ignores the actual capybara webkit code and would be quite difficult to actually do an integration test.

The problem is currently that the test will always return true no matter what because no error is raised.

mikepack:
Got it. So, to clarify, you have a test that should fail, but is passing? This test is passing because the evaluation script does not rerun?

mikepack:
Or, is the result of the test not being captured? It's neither passing nor failing. More like skipped. This is what I'd expect, not that the test actually passes.

ryanong:
more like skipped. The functionality that would make this fail is stubbed out and would only really fail if we added a 5 second sleep of some sort.

mikepack:
Thanks Ryan!

PR #406, opened 2015-09-08
Add .es6.js to normalize_js_extension

bouk:
@jejacks0n @mikepack 

Some people use this as an extension for es6 files.

This might be better off as a configuration parameter, or alternatively you could look at `Sprockets.mime_types["text/ecmascript-6"]` to find out which extensions are registered to that mime type

### DIFF
--- a/lib/teaspoon/suite.rb
+++ b/lib/teaspoon/suite.rb
@@ -95,7 +95,7 @@ module Teaspoon
     end
 
     def normalize_js_extension(filename)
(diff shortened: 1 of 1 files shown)


jejacks0n:
Can you confirm that the breakage isn't from your addition? It doesn't appear to be directly tied, but would appreciate your time.

bouk:
@jejacks0n seemed to have been a fluke

jejacks0n:
Thanks!

PR #407, opened 2015-09-11
Removed outdated css prefixes

iainbeeston:
I use [autoprefixer](https://github.com/postcss/autoprefixer) in my rails project (it automatically adds vendor specific prefixes to css rules when necessary). However, it's reporting that some of the css in teaspoon is using invalid css for linear gradients:

```
autoprefixer: /Users/iain/.gem/ruby/2.2.3/gems/teaspoon-1.0.2/app/assets/stylesheets/teaspoon.css:124:3: Gradient has outdated direction syntax. New syntax is like "to left" instead of "right".
```

Rather than correct the syntax for gradients, instead I've removed all css vendor prefixes that are no longer required by current browsers, based on [caniuse](http://caniuse.com). The only prefixed rule left is `-webkit-transform`, as apparently current versions of safari (version 8.x) still require the prefix.

### DIFF
--- a/app/assets/stylesheets/teaspoon.css
+++ b/app/assets/stylesheets/teaspoon.css
@@ -117,13 +117,7 @@ html[xmlns] #teaspoon-interface .teaspoon-clearfix {
   width: 87px;
   height: 34px;
   -webkit-transform: rotate(-18deg);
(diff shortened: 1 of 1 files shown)


mikepack:
Thanks Iain!

PR #418, opened 2015-10-07
Configurable JavaScript extensions

mikepack:
In lieu of #391.

@jejacks0n :eyes: please :heart:

### DIFF
--- a/lib/teaspoon/suite.rb
+++ b/lib/teaspoon/suite.rb
@@ -94,8 +94,10 @@ module Teaspoon
       normalize_js_extension(filename)
     end
 
-    def normalize_js_extension(filename)
-      filename.gsub(".erb", "").gsub(/(\.js\.coffee|\.coffee|\.es6|\.js\.es6|\.es6\.js)$/, ".js")
+    def normalize_js_extension(original_filename)
+      config.js_extensions.inject(original_filename.gsub(".erb", "")) do |filename, extension|
+        filename.gsub(Regexp.new(extension.to_s + "$"), ".js")
+      end
     end
 
     def glob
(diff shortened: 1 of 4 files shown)


lib/teaspoon/suite.rb
```diff
@@ -94,8 +94,10 @@ def asset_from_file(original)
       normalize_js_extension(filename)
     end
 
-    def normalize_js_extension(filename)
-      filename.gsub(".erb", "").gsub(/(\.js\.coffee|\.coffee|\.es6|\.js\.es6|\.es6\.js)$/, ".js")
+    def normalize_js_extension(original_filename)
```
mikepack: @jejacks0n I'll look tomorrow but do you know offhand if sprockets can do this normalization for us? Would be nice.

jejacks0n: If I recall it was not public and involved a bunch of complexity around tilt or something to determine what extensions are understood, and I think maybe wanted to pass the file off at that point to be processed. So not super useful there, but I'm not 100% sure. The config is only to try and find spec files (if I recall), and that's through the file system via a glob call. Might be worth code diving to see if that's improved since last time though.

---
Jeremy Jackson

jejacks0n:
Looks good to me. Side note about documenting in the wiki so it's captured there.

hlascelles:
A note to anyone coming to this PR wondering why their `.js.es6` specs are not running... You have to set  your matcher to include them.

    suite.matcher = '{spec/javascripts,app/assets}/**/*_spec.{js,js.es6,js.coffee,coffee}'

@jejacks0n, I'd say this may be a bug, or at least unexpected. If you look in *configuration.rb:69* you see

    @matcher       = "{spec/javascripts,app/assets}/**/*_spec.{js,js.coffee,coffee}"

But that disagrees with *configuration.rb:79*

    @js_extensions = [/(\.js)?\.coffee/, /(\.js)?\.es6/, ".es6.js"]

One supports `es6`, and one doesn't. Shouldn't they both match the same suffixes?

jejacks0n:
happy to get a PR.

hlascelles:
Great, just checking I hadn't missed anything. Created: https://github.com/modeset/teaspoon/pull/488

PR #420, opened 2015-10-13
Add an option to specify the server host

### LINKED ISSUE #397: Error after upgrading to 1.0.x: Can't find variable: Teaspoon
upgrading from 0.9.1 to 1.0.2 gives this error:

```
Thin web server (v1.6.2 codename Doc Brown)
Maximum connections set to 1024
Listening on 0.0.0.0:5000, CTRL+C to stop
Teaspoon running foobar suite at http://127.0.0.1:5000/teaspoon/foobar
ReferenceError: Can't find variable: Teaspoon
  # http://127.0.0.1:5000/teaspoon/foobar?reporter=Console:14

ReferenceError: Can't find variable: Teaspoon
  # http://127.0.0.1:5000/teaspoon/foobar?reporter=Console:143
```

I wish I could provide more info, but I am not sure how/where to get a stacktrace of the error, also not sure if there is a verbose mode or something that could give me more info to debug the problem.

thanks

rafaelfranca:
By default teaspoon starts the server on localhost interface. This
breaks some setup that use different hosts like vagrant.

We are introducing a new option so the server host can be changed.

Closes #397.

cc @mikepack

### DIFF
--- a/lib/teaspoon/configuration.rb
+++ b/lib/teaspoon/configuration.rb
@@ -25,13 +25,14 @@ module Teaspoon
 
     # console runner specific
 
-    cattr_accessor :driver, :driver_options, :driver_timeout, :server, :server_port, :server_timeout, :fail_fast,
+    cattr_accessor :driver, :driver_options, :driver_timeout, :server, :server_host, :server_port, :server_timeout, :fail_fast,
                    :formatters, :color, :suppress_log,
                    :use_coverage
     @@driver         = Teaspoon::Driver.default
     @@driver_options = nil
     @@driver_timeout = 180
     @@server         = nil
+    @@server_host    = nil
     @@server_port    = nil
     @@server_timeout = 20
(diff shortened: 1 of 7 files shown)


lib/teaspoon/configuration.rb
```diff
@@ -25,13 +25,14 @@ class Configuration
 
     # console runner specific
 
-    cattr_accessor :driver, :driver_options, :driver_timeout, :server, :server_port, :server_timeout, :fail_fast,
+    cattr_accessor :driver, :driver_options, :driver_timeout, :server, :server_host, :server_port, :server_timeout, :fail_fast,
```
rafaelfranca: Sorry @houndci, but splitting this line just to please style will harm the git history of the line below. I prefer the keep the git history as clean as possible so changes are easier to find when tracing a bug.

jejacks0n:
Thanks! :+1:

PR #421, opened 2015-10-13
Fix base formatter when there are no params on the filename

rafaelfranca:
In my project the body and instrumnet query params were not being added to the file name I believe because we use the already compiled assets to run the tests. Because of this the base formatter were failing because it expects the query parameters to exists.


mikepack:
Thanks Rafael!

PR #422, opened 2015-10-14
Use jQuery instead of $ to check jQuery existence

rafaelfranca:
When `jQuery.noConflict()` is called the `$` can be something different than `jQuery` like the webkit's default selector helper.

### DIFF
--- a/app/assets/javascripts/teaspoon/fixture.coffee
+++ b/app/assets/javascripts/teaspoon/fixture.coffee
@@ -2,7 +2,7 @@ class Teaspoon.Fixture
 
   @cache: {}
   @el: null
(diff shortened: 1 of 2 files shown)


jejacks0n:
This was technically to support some of the micro libraries that use $, but it's worth merging to see if it impacts anybody. Cheers!

rafaelfranca:
I see. Maybe we can check the presence of `parseHTML` instead? My test suite is failing because `parseHTML` is `undefined` when `$` is the browser selector helper.

jejacks0n:
Nah, I think this is good. :+1:

rafaelfranca:
Yeah, Zepto doesn't have `parseHTML` so it will fail anyway.

PR #423, opened 2015-10-14
Use faster travis infrastructure

rafaelfranca:
See http://docs.travis-ci.com/user/migrating-from-legacy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: false
+cache: bundler
 rvm:
   - ruby-1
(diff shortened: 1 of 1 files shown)


rafaelfranca:
This speed up the tests matrix in almost 10 minutes.

Before: https://travis-ci.org/modeset/teaspoon/builds/85185407
After: https://travis-ci.org/modeset/teaspoon/builds/85259893

jejacks0n:
That is faster. Haha.

PR #424, opened 2015-10-14
Build static releases for the Fixture fix

rafaelfranca:
I didn't know that it was required when I opened #422. Without this my change don't apply to my application.


jejacks0n:
Ah, so you can switch to the non-static files if you want (in your config), but it requires coffeescript and takes a bit longer to process.

PR #439, opened 2015-12-06
Skip to raise ::DependencyError in phantomjs#run

speed-of-light:
if given block returns nil, $?.success will produce
```
  undefined method `success?' for nil:NilClass (NoMethodError)
```

given command:
```
teaspoon -C default
```


mikepack:
Under what cases will `$?` be nil? Can you explain how to reproduce this, or write a spec? Are you missing a phantomjs binary?

PR #449, opened 2016-01-18
Rename Registry#equal? to Registry#matches?

### LINKED ISSUE #448: Teaspoon::Registry#equal? is incompatible with Object#equal?
Teaspoon::Registry defines an `equal?` instance method with two arguments in https://github.com/modeset/teaspoon/blob/cdf46b24676c093f7a2b0ed8e8e473dd89a47564/lib/teaspoon/registry.rb#L29. This signature is incompatible with `Object#equal?` which only takes a single argument. The rdoc even mentions the following:

```
Unlike ==, the equal? method should never be overridden by subclasses as it is
used to determine object identity (that is, a.equal?(b) if and only if a is
the same object as b):
```

This in turn breaks Rubinius as described in https://github.com/rubinius/rubinius/issues/3570 as we use `#equal?` here: https://github.com/rubinius/rubinius/blob/34af1dbb3c24aab455df12e4ce1c0568f1110d47/kernel/common/binding.rb#L14

There's only a single instance of this `equal?` method being used in the Teaspoon code, so renaming the method shouldn't be too much of a problem: https://github.com/modeset/teaspoon/blob/70fba0054a1155c326a120b99ff26934e97330ab/lib/teaspoon/engine.rb#L83

YorickPeterse:
This ensures the method doesn't re-define Object#equal? in a way that's
not compatible with the default implementation.

Fixes #448

### DIFF
--- a/lib/teaspoon/engine.rb
+++ b/lib/teaspoon/engine.rb
@@ -80,7 +80,7 @@ module Teaspoon
       private
 
       def self.using_phantomjs?
-        Teaspoon::Driver.equal?(Teaspoon.configuration.driver, :phantomjs)
+        Teaspoon::Driver.matches?(Teaspoon.configuration.driver, :phantomjs)
       end
 
       def self.render_exceptions_with_javascript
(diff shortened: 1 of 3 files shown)


YorickPeterse:
Any comments on this? Is this good to go?

jejacks0n:
Thanks. =)

PR #452, opened 2016-02-04
javascript_include_tag override is also needed for Rails 5

jmbejar:
We're testing a Rails 5 app using the `rails 5` branch. The only thing that we need to adjust is the Rails version comparison used to check if the `javascript_include_tag` method should be overridden.

### DIFF
--- a/lib/teaspoon/engine.rb
+++ b/lib/teaspoon/engine.rb
@@ -98,7 +98,7 @@ end
 
 begin
   require 'action_view'
-  if ActionView::VERSION::STRING.start_with?('4.2.5')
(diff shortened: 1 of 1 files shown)


lib/teaspoon/engine.rb
```diff
@@ -98,7 +98,7 @@ def render_exception(_env, exception)
 
 begin
   require 'action_view'
-  if ActionView::VERSION::STRING.start_with?('4.2.5')
+  if Rails.version >= '4.2.5'
```
jmbejar: Will do

jejacks0n:
Cheers!

PR #453, opened 2016-02-12
Add jasmine 2.4.1 to teaspoon-jasmine

ryands:
Adds jasmine 2.4.1 to teaspoon-jasmine.


jejacks0n:
Thanks! Yeah, this is why we broke them off into separate "sub" gems.. at some point later we'll probably extract them and have different repos, but this makes life easier for now. I'll merge, leave this on master for a bit, and then release it after some testing has happened.

PR #454, opened 2016-02-16
Use Phantomjs gem if it's defined

BrianHawley:
If the Phantomjs gem is installed, it should be used first. The first
thing it checks is the equivalent to the `which` function, but with
version checking. It never returns a blank value.

### DIFF
--- a/lib/teaspoon/driver/phantomjs.rb
+++ b/lib/teaspoon/driver/phantomjs.rb
@@ -54,8 +54,7 @@ module Teaspoon
 
       def executable
         return @executable if @executable
-        @executable = which("phantomjs")
-        @executable = ::Phantomjs.path if @executable.blank? && defined?(::Phantomjs)
+        @executable = defined?(::Phantomjs) ? ::Phantomjs.path : which("phantomjs")
         return @executable unless @executable.blank?
         raise Teaspoon::MissingDependencyError.new("Unable to locate phantomjs. Install it or use the phantomjs gem.")
       end


jejacks0n:
Can you give me some background? I specifically took this out a long while back because it wasn't very good with OS level support, and I didn't want it as a dependency.

BrianHawley:
If you're using the Phantomjs gem, you are generally using it on purpose. It declares a specific version of the browser for compatibility, and installs that version when it isn't the one installed in the system.

If you don't want to use it (perhaps because you aren't using Linux or OSX), you don't have to. In that case `Phantomjs` won't be defined, and the `which` function will be called instead.

The only difference this change makes is to assume that if the developer has decided to use the gem, that their wishes should be respected, rather than trying the `which` function first.

BrianHawley:
Pushing up a new version without the style complaints.

BrianHawley:
We tend to use a few different testing environments around here, so we've found the phantomjs gem to be useful for managing compatibility with specific versions, especially when upgrades have to be tied to corresponding test changes. With the priorities set in teaspoon now, `Phantomjs.path` never gets called if there is a preinstalled version, even if that version doesn't match what you need.

We're evaluating switching over to using teaspoon, but not being able to use the gem is a deal breaker (the only one I can see so far though).

I've also been looking into fixing some of the cross-platform compatibility issues for mentioned, but that's an issue for another day.

jejacks0n:
Can you please update the wiki or readme based on where you would like this to be documented please?

BrianHawley:
Sure, will do.

PR #455, opened 2016-02-17
Fix the usage of `time` in the junit formatter

sguha00:
`timestamp` should be used to record when the test ran, `time` is the time taken to execute the tests

https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd#L195 - `time`
https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd#L165 - `timestamp`


mikepack:
Thanks!

sguha00:
Sorry, I `start` is actually a `String` not a `Time` object. I made another pull request https://github.com/modeset/teaspoon/pull/456 to fix it

PR #468, opened 2016-03-23
Fix support for Sprockets 4

liseki:
``Sprockets::Index`` was a reference to the cached environment in Sprockets 2, deprecated in Sprockets 3, and removed in Sprockets 4.

### DIFF
--- a/lib/teaspoon/engine.rb
+++ b/lib/teaspoon/engine.rb
@@ -51,6 +51,9 @@ module Teaspoon
 
     def self.inject_instrumentation
       Sprockets::Environment.send(:include, Teaspoon::SprocketsInstrumentation)
+      Sprockets::CachedEnvironment.send(:include, Teaspoon::SprocketsInstrumentation)
+    rescue NameError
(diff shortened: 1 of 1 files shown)


lib/teaspoon/engine.rb
```diff
@@ -51,6 +51,9 @@ def self.add_precompiled_assets(assets)
 
     def self.inject_instrumentation
       Sprockets::Environment.send(:include, Teaspoon::SprocketsInstrumentation)
+      Sprockets::CachedEnvironment.send(:include, Teaspoon::SprocketsInstrumentation)
+    rescue NameError
```
mfazekas: I think it would be better if the rescue was only for Sprockets::CachedEnvironment and not for Sprockets::Environment. We might want to use ```defined? Sprockets::CachedEnvironment```

liseki:
@mfazekas Sorry I was not able to see your comment before the commit happened. What you suggest would have been a bit more nuanced.

PR #469, opened 2016-03-29
Extended runner.js to provide further error details

sebastianzillessen:
If there has been an error on the teaspoon webpage (e.g. if the Spec helper tries to require a file/folder which does not exist for some reason as an empty folder that has been removed during a CI process) which will result in a 'success' return code, but does not render the Teaspoon test page, the error description was - especially while debugging on a remote CI - not really helpful.

This commit provides additional error descriptions in this case by rendering the plain_text of the returned website into the console output, it does not modify the functionality.

### DIFF
--- a/lib/teaspoon/driver/phantomjs/runner.js
+++ b/lib/teaspoon/driver/phantomjs/runner.js
@@ -89,8 +89,32 @@
             return window.Teaspoon;
           });
           if (!(status === "success" && defined)) {
-            _this.fail("Failed to load: " + _this.url);
-            return;
+            if (status === "success") {
+              // Could not load the window.Teaspoon object from the JavaScript on the
+              // rendered page. This indicates that a problem occured. Lets therfore
+              // print the page as a failure description.
+              // Get plain text of the page, intend all lines (better readable)
+              var ind = "   ";
+              var error = _this.page.plainText.replace(/(?:\n)/g, "\n" + ind);
+              // take only first 10 lines, as they usually provide a good entry
+              // point for debugging and we should not spam our console.
+              var erroroutput = error.split("\n").slice(0, 10);
+              if (erroroutput !== error) {
+                erroroutput.push("... (further lines have been removed)");
+              }
+              var fail = [
+                "Failed to get Teaspoon result object on page: " + _this.url,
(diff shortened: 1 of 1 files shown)


jejacks0n:
I'll keep this around, but it probably won't be merged as is. So initially I think Teaspoon handled these things well, by telling Rails to convert the rendered output to a javascript exception, but that seems to not work anymore. I'd rather handle this at the suite_controller level and handle exceptions by rendering that exception as a "<script>throw [exception]<script>" sort of statement.  That's probably a bit more elegant and robust.  But thanks for bringing it to our attention -- it definitely needs some love there.

Worth noting, some gems, like better_errors can make this stuff not work. Are you using anything like that?

sebastianzillessen:
Well, the thing is, it took me almost 3 hours to find the problem in my test suite, because teaspoon just tells u, it could not connect to the url ...:xx/teaspoon/suite if there exists any error on the page. 
The problem for me was, that I used a `require_tree` in my spec_helper, which was (as it was empty) ignored by the CI while cloning the repository and therefore my specs failed. 

So I think it could safe other people time if they are just using the console or a CI environment, by providing them better error explanations.

And as it is only a problem in the Problem in the `phantomjs/runner` I think this is the right place for it.

Yes, we are indeed using better_errors, therefore I thought it would be more reliable to check it here as well, as the approach I took tackles the actual problem, namely that `window.Teaspoon` is not present in the page returned to phantomjs.

So to sum it up: I think we should include it, so that everyone sees error messages, if something goes wrong with the website while the tests are running.

jejacks0n:
That's fair. I understand what you're trying to solve. There's likely a better way to do this, and I'd like to investigate that before considering your solution. Also, there aren't any specs here and it seems to have broken the build, but instead of saying that, I figured it might be worth looking into alternate approaches to help you resolve the problem without asking you for additional work.

I'm sorry Teaspoon fell down on you here, but it's really about how better_errors hooks in (or doesn't really hook in, but tends to stomp all over) how rails handles exceptions normally. That seems like a better thing to look into from my perspective.

sebastianzillessen:
Alright, always glad to help. Let me just know, how you fixed it. :+1:

Issue #470, opened 2016-03-29
JUnit Formatter overwrites xml report when multiple test suites are in place


sebastianzillessen:
I have multiple test suites, as my application has different main layouts with different javascript assets.

My configuration looks like this:

```
  config.suite do |suite|
    suite.matcher = "**/application/**/*_spec.{js,js.coffee,coffee}"
    suite.helper = File.expand_path("../javascripts/spec_helper_application.js", __FILE__)
    ...
  end
  # new config to run admin tests
  config.suite :admin do |suite|
    suite.matcher = "**/admin/**/*_spec.{js,js.coffee,coffee}"
    suite.helper = File.expand_path("../javascripts/spec_helper_admin.js", __FILE__)
  end
  FileUtils.mkdir_p(dirname = Rails.root.join("spec/reports"))
  config.formatters = ["dot", "junit>#{dirname.join("teaspoon.xml")}"]
```

The important part are the formatters. Having a look at `teaspoon/lib/teaspoon/runner.rb#24` the `Runner` uses the same output file for each test-suite. In `teaspoon/lib/teaspoon/formatter/base.rb#21` the `Formatter::Base` defines, that it will truncate existing files to zero length (`"w":  Write-only, truncates existing file to zero length or creates a new file for writing.`). Hence, you will never see the output of all test-suites but only the last one. 

Is there any configuration I missed, that could help me with that issue?

mathieujobin:
looks like the patch for this was merged
thank you

PR #471, opened 2016-03-29
#470: providing option to use placeholders in output file for formatters

sebastianzillessen:
When defining multiple suites and running the tests on a CI (like Jenkins), you might need the output of `bundle exec teaspoon` in your XML-reports.
To get one XML-file per suite you are running, you might want to add e.g a `junit` formatter which creates one result file per suite. 
 
```ruby
config.formatters = ["junit>#{Rails.root.join(["spec", "reports", "teaspoon_%{suite_name}.xml"])}"]
``` 

Will create a single file for each defined test-suite under `spec/reports/teasoon_%{suite_name}.xml`. 
Allowed placeholders are:

- `suite_name`: the name defined by your config (`config.suite :my_suite ...`)
- `date`: the execution timestamp of your test (`Date.now.to_i`)

### DIFF
--- a/lib/teaspoon/formatter/base.rb
+++ b/lib/teaspoon/formatter/base.rb
@@ -7,7 +7,7 @@ module Teaspoon
 
       def initialize(suite_name = :default, output_file = nil)
         @suite_name  = suite_name.to_s
-        @output_file = output_file
+        @output_file = parse_output_file(output_file)
         @stdout      = ""
         @suite       = nil
         @last_suite  = nil
@@ -153,6 +153,24 @@ module Teaspoon
         filename = uri.path.sub(%r(^/assets/), "")
         filename += "?#{params.join("&")}" if params.any?
         filename
+      end
+
+      def parse_output_file(output)
+        return output unless output
+        output.gsub(/%{([^}]*)}/) { parse_output_capture($1) }
+      end
+
+      def parse_output_capture(cap)
+        case cap
+        when "suite_name"
+          @suite_name
+        when "date"
+          Time.now.to_i
+        else
+          warn ["Teaspoon::Formatter - Output File can only contain the placeholders %{suite_name} or %{date}.",
+                "%{#{cap}} is not supported and will be ignored."].join("\n")
+          ""
+        end
       end
     end
   end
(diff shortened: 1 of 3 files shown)


lib/teaspoon/formatter/base.rb
```diff
@@ -7,7 +7,7 @@ class Base
 
       def initialize(suite_name = :default, output_file = nil)
         @suite_name  = suite_name.to_s
-        @output_file = output_file
+        @output_file = parse_output_file output_file
```
jejacks0n: Let's use parens for method calls.

lib/teaspoon/formatter/base.rb
```diff
@@ -154,6 +154,25 @@ def filename(file)
         filename += "?#{params.join("&")}" if params.any?
         filename
       end
+
+      def parse_output_file(output)
+        return output unless output
+        match = output.scan(/%{([^}]*)}/)
```
jejacks0n: Any reason to use scan instead of gsub with a proc?

```
output.gsub!(/%{([^}]*)}/) { |cap| parse_output_capture(cap) }
```

Then you can extract the switch (aka case) statement into a separate method because I'm not a fan of the indentation style. =)

lib/teaspoon/formatter/base.rb
```diff
@@ -154,6 +154,25 @@ def filename(file)
...
+        match.flatten.each do |m|
+          rep = case m
+                when "suite_name"
+                  @suite_name.underscore
+                when "date"
+                  Date.now.to_i
+                else
+                  warn "Teaspoon::Formatter - Output File can only contain the placeholders %{suite_name} or %{date}."
```
jejacks0n: I had to look up warn, cause I've never used it. Interesting, and probably a good use for it.

jejacks0n:
This looks good, and thank you for spending the time on it.

I added some comments for changes. And I'd like to see a new spec.

jejacks0n:
Thanks for resolving that, and spending the time on it. :+1:

sebastianzillessen:
Hi @jejacks0n unfortunately the PR has not resulted in a new version of teaspoon. When will u generate a new one so that we can use this feature?
Thanks!

sebastianzillessen:
Hey @jejacks0n When can we expect a new version with this feature included?

Issue #428, opened 2015-11-05
console.log not printed to stdout when using capybara-webkit


volkergersabeck:
using the capybara-webkit driver I was not able to see any output from console.log statements in my javascript on the stdout. switching to phantomjs showed the output immediately.

my configuration is the default:
#config.suppress_log = false

from my Gemfile.lock:
capybara-webkit (1.7.1)
teaspoon (1.1.1)
teaspoon-jasmine (2.3.4)

kprinssu:
You'll need to run the webserver via ```rails s``` and go to ```http://localhost:3000/teaspoon``` to get console output.

jejacks0n:
If this isn't working it should probably be -- if for nothing other than parity. If there's any time available we'll look into it.. thanks for the heads up.

starsirius:
+1

jejacks0n:
Please try that branch and report back. I have only tested it on the latest capybara-webkit (1.10.1 as of writing) and it appears to work -- no idea about previous versions.

starsirius:
@jejacks0n Thanks for taking a look. I've tried using the above branch, but now it looks like not running the specs and just exits. Will try to take a closer look, but for your information, this is all I got now:
```
$ RAILS_ENV=test bundle exec teaspoon
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2-compliant syntax, but you are running 2.3.0.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
I, [2016-04-11T16:14:28.244418 #11913]  INFO -- : ** [Raven] Raven 0.13.2 configured not to send errors.
Starting the Teaspoon server...
Teaspoon running default suite at http://127.0.0.1:54923/teaspoon/default


$ 
```

jejacks0n:
Thanks @starsirius, that's likely because of the capybara-webkit version (I think).. can you make sure you're using 1.10.1 for me as a follow up? That layer changed a lot between versions of capybara-webkit, and so I expected some hassles there. Please provide the version you're using, and what the results are with 1.10.1 if you have the time.

starsirius:
@jejacks0n Confirmed that using capybara-webkit 1.10.1 works. It's running the tests and outputting the `console.log`s as long as outputs from other libraries! 👍  For example, the warning from React:
> Warning: React.render is deprecated. Please use ReactDOM.render from require('react-dom') instead.

[...]

PR #472, opened 2016-04-08
Attempts to resolve the capybara webkit STDOUT issues. #428

jejacks0n:
This leaves a bit to be desired, but is really the only way I could figure out how to accomplish this after teasing apart capybara/capybara webkit.

The main issue I have with it is the order in which things are output.. if you add console.log inside your first spec, it will behave as expected.. if you put another console.log inside your second spec, that output is visible _before_ the dot from the first spec. I assume this is the nature of the speed at which capybara is running / reading from the socket, but it could lead to confusion.


jejacks0n:
If anyone from the issue would like to dive into this further, it might be useful to get more eyes on it.

PR #473, opened 2016-04-11
Add support for Sprockets 4

liseki:
[Fix support for Sprockets 4](https://github.com/modeset/teaspoon/pull/468) ensured that Teaspoon successfully loaded with Sprockets 4. This finishes up the job to ensure Teaspoon works with Sprockets 4.

### DIFF
--- a/lib/teaspoon/instrumentation.rb
+++ b/lib/teaspoon/instrumentation.rb
@@ -45,14 +45,14 @@ module Teaspoon
 
     def self.ignored?(asset)
       Array(Teaspoon::Coverage.configuration.ignore).any? do |ignore|
-        asset.pathname.to_s.match(ignore)
+        asset.filename.match(ignore)
       end
     rescue Teaspoon::UnknownCoverage
       false
     end
 
     def add_instrumentation(asset)
-      source_path = asset.pathname.to_s
+      source_path = asset.filename
       Dir.mktmpdir do |temp_path|
         input_path = File.join(temp_path, File.basename(source_path)).sub(/\.js.+/, ".js")
         File.open(input_path, "w") { |f| f.write(asset.source) }
--- a/lib/teaspoon/suite.rb
+++ b/lib/teaspoon/suite.rb
@@ -53,8 +53,7 @@ module Teaspoon
         asset = @env.find_asset(source)
 
         if asset && asset.respond_to?(:logical_path)
-          assets = config.expand_assets ? asset.to_a : [asset]
-          assets.map { |a| asset_url(a) }
+          asset_and_dependencies(asset).map { |a| asset_url(a) }
         else
           source.blank? ? [] : [source]
         end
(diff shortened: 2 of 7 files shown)


lib/teaspoon/instrumentation.rb
```diff
@@ -45,14 +45,14 @@ def instrumented_response
 
     def self.ignored?(asset)
       Array(Teaspoon::Coverage.configuration.ignore).any? do |ignore|
-        asset.pathname.to_s.match(ignore)
+        asset.filename.match(ignore)
```
jejacks0n: Is `filename` present in current versions, or is this a breaking change for others?

lib/teaspoon/suite.rb
```diff
@@ -53,7 +53,16 @@ def asset_tree(sources)
         asset = @env.find_asset(source)
 
         if asset && asset.respond_to?(:logical_path)
-          assets = config.expand_assets ? asset.to_a : [asset]
+          assets =
```
jejacks0n: Might want to move this to a different method that can be easily overridden if the case comes up.

spec/javascripts/spec_helper.js
```diff
@@ -0,0 +1,11 @@
+//= require support/json2
```
jejacks0n: Was this changed to a .js file because you don't have coffeescript?

jejacks0n:
Thanks! Generally looks good, but I had some questions.

lib/teaspoon/instrumentation.rb
liseki: ```filename``` is present in Sprockets 3 & 4 and not in 2. I patched Sprockets 2 to support it in ```teaspoon/engine.rb```

lib/teaspoon/suite.rb
liseki: Will do.

lib/teaspoon/instrumentation.rb
jejacks0n: :+1:

spec/javascripts/spec_helper.js
liseki: In Sprockets 4 the top level/wrapper asset sets the MIME type that all its included assets will have. If we use ```spec_helper.coffee``` any attempt to include a JS file (such as ```support/json2```) will fail since Sprockets can not convert JS to CoffeeScript. [See this comment] (https://github.com/rails/sprockets/issues/273#issuecomment-206415863) for a slightly longer explanation.

liseki:
I did locally run the specs with ```appraisal``` and this works for Sprockets 2, 3, and 4 so there should be no back support breakage. I presume the CI also does this.

jejacks0n:
In theory it does, but I wasn't certain about the version of sprockets in the Rails 3 build. Thanks for clarifying.

spec/javascripts/spec_helper.js
jejacks0n: It may not be a concern of this PR then, but that's a fairly large change in behavior that will definitely impact a lot of users of Teaspoon / the asset pipeline in general. Ok, I'll create an issue to track that since we'll need to adjust the install generator to support that. If you want to tackle that it would be awesome too, but don't feel obligated since it's only tangentially related.

liseki: I'm happy to work on that!

jejacks0n: Awesome, thanks again! Will merge when you give approval.

Issue #475, opened 2016-04-12
#generate_report sometimes fails with NoMethodError: undefined method `exitstatus' for nil:NilClass


baburdick:
```
E, [2016-04-11T21:16:47.045563 #74293] ERROR -- : reaped #<Process::Status: pid 74318 SIGSEGV (signal 11)> worker=unknown
rake aborted!
NoMethodError: undefined method `exitstatus' for nil:NilClass
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/coverage.rb:37:in `block in check_thresholds'
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/coverage.rb:54:in `call'
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/coverage.rb:54:in `block in input_path'
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/coverage.rb:51:in `input_path'
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/coverage.rb:35:in `check_thresholds'
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/runner.rb:63:in `resolve_coverage'
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/runner.rb:53:in `on_result'
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/runner.rb:30:in `notify_formatters'
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/runner.rb:16:in `process'
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/driver/phantomjs.rb:28:in `block in run_specs'
/Users/me/.rvm/gems/ruby-2.2.4@my_app/gems/teaspoon-1.1.5/lib/teaspoon/driver/phantomjs.rb:35:in `each'
[...]

PR #476, opened 2016-04-13
Use Open3.capture2e instead of %x{} and $?

baburdick:
note: this fixes text + text-summary coverage reports
      (previously generated only text report when both were config'd --
      and the spec enshrined this)

note: using `$?` with `%x{}` can lead to missing exit status code issues

ref: modeset/teaspoon#475


baburdick:
Fixing hound-bot critiques ...

baburdick:
@jejacks0n: any thoughts on this PR?

jejacks0n:
It's looking good -- thanks for the time you've put into it. Can you provide a link so I understand why `$?` with `%x{}` can lead to missing exit status code issues? Just want to learn from it.  Also, in a few other places I use `IO.popen` -- is there a reason not to do the same here? Again, mostly just trying to figure out the subtle differences.

baburdick:
I wish I could provide a link. And I wish I understood better what's at the core of the problem. But it's a conclusion I've drawn from this experience. On my machine, using pry, I was able to witness this issue, described in modeset/teaspoon#475 .

Using this solution, which returns STDOUT, STDERR, and the exit status code in a single array, I can't reproduce the issue.

With respect to `IO.popen`, I see that you're using it in these two files, and depending on Ruby on subsequent lines to put `$?` into the local scope:
* `lib/teaspoon/driver/phantomjs.rb`
* `lib/teaspoon-devkit.rb`

If `IO.popen` uses the same mechanism to put `$?` into the local scope, the same problem could manifest. But I haven't seen it in my environment. And so far I haven't seen anyone else report a similar issue (although my Google-foo may have failed me in that regard).

Using the standard library's `Open3` and one of its `capture*` methods to return the subset of STDOUT, STDERR, and exit status code that you need definitely eliminates depending on Ruby's mechanism to insert the exit status code in to the local scope in `$?`.

The only thing I can further suggest is to take a look at my change to the "generates coverage reports" spec. If you agree that it is now testing for the proper output, given the inputs, then something is not working properly in master. But I'm not certain the cause is what's described above. My changes expose more of the implementation to testability. So that fix could just be a side-effect of the better testability.

I wish I had an open-and-shut case that you could reproduce, or could point definitively to an existing bug in Ruby. But I have to depend on your judgement here. I hope my reasoning is clear and sensible to you.

baburdick:
Checking back in. It looks like [`Open3`](http://ruby-doc.org/stdlib-2.2.1/libdoc/open3/rdoc/Open3.html#method-c-popen3) provides alternatives to [`IO.popen`](http://ruby-doc.org/core-2.2.1/IO.html#method-c-popen): [`Open3.popen2`](http://ruby-doc.org/stdlib-2.2.1/libdoc/open3/rdoc/Open3.html#method-c-popen2), [`Open3.popen2e`](http://ruby-doc.org/stdlib-2.2.1/libdoc/open3/rdoc/Open3.html#method-c-popen2e), and [`Open3.popen3`](http://ruby-doc.org/stdlib-2.2.1/libdoc/open3/rdoc/Open3.html#method-c-popen3). For your use of `IO.popen` and `$?`, it looks like you could substitute `Open3.popen2e`.

Also, here's a few links providing more insight into what the various forms of process spawning do and how and why:
* http://unknownerror.org/opensource/ruby/ruby/q/stackoverflow/690151/getting-output-of-system-calls-in-ruby (Hat-tip: @albertosaurus)
* http://stackoverflow.com/questions/7212573/when-to-use-each-method-of-launching-a-subprocess-in-ruby
* http://unethicalblogger.com/2011/11/12/popen-can-suck-it.html (an older article, principally germane to Ruby 1.8 and 1.9, that sheds some light on the mingling of STDERR in the parent and child processes that does NOT happen in Open3.)

baburdick:
@jejacks0n: You may have missed my last post here. Please take a look. It may satisfy your curiosity.

ivdma:
@jejacks0n this still isn't merged :(

patrickmclaren:
Anything left for this to merge?

baburdick:
@jejacks0n: Just checking in. Completely understood if you're very busy. Looking forward to when you can check this out and let me know if there's more you need from me. (Happy New Year, BTW.)

baburdick:
Not sure what's breaking the build. It appears that merging in the most recent master and pushing caused current Travis to expose issues in most recent yet old master.

baburdick:
@jejacks0n: Looks like jejacks0n/teaspoon#534 fixes all the breakage here. Let me know what I can do further here.

baburdick:
Thank you, @jejacks0n!

baburdick:
Verified that this works. All that remains is to build and push to rubygems.

Issue #479, opened 2016-05-05
Errors outside of tests don't cause the test run to fail


bouk:
This seems to have been discussed before in previous issues, but we've ran into multiple problems at Shopify where something in our javascript bundle throw a SyntaxError or ReferenceError or whatever, which would then not make Teaspoon fail and instead cause our CI to silently succeed and us to miss it.

I propose that we make Teaspoon fail the test run if there's an error before `Teaspoon.execute` is ran, as that will make it non-disrupting when there's errors in asynchronous things, but still makes sure this doesn't silently let errors slip through.

bouk:
previous issues; https://github.com/modeset/teaspoon/issues/445 https://github.com/modeset/teaspoon/issues/425

jejacks0n:
Yeah, I've been torn on this because it doesn't indicate a real failure directly. That being said, it's an easy fix with phantomjs, but may not be true elsewhere.

If you decide to look into this I tend to think we might need to do this at the controller level (suite_controller) and handle that exception by ending everything.

---
Jeremy Jackson

bouk:
fixed

PR #480, opened 2016-05-05
Add onerror handler for pretest errors

bouk:
@jejacks0n this is not meant for merging yet but more to discuss different approaches for #479 

The approach I took here will work for any driver, but has the drawback of showing very limited information. `window.onerror` doesn't give a whole lot of info. This does work however, and my test fails as expected when the error happens

I'll also need some help with how/where to define a test for this.


teaspoon-jasmine/lib/teaspoon/jasmine/assets/teaspoon-jasmine1.js
```diff
@@ -77,12 +95,12 @@
...
-      Teaspoon.messages.push(arguments[0]);
+      var e, error;
+      Teaspoon.messages.push(msg);
       try {
         return typeof console !== "undefined" && console !== null ? typeof console.log === "function" ? console.log.apply(console, arguments) : void 0 : void 0;
-      } catch (_error) {
-        e = _error;
+      } catch (error) {
+        e = error;
         throw new Error("Unable to use console.log for logging");
```
jejacks0n: I'd probably name these `error_`, or similar to handle the ie8 issue.

jejacks0n: Nevermind, this is just coffeescript -> javascript.

jejacks0n:
Hey @bouk, thanks for taking this on, and sorry I've been absent from it a bit. I'll merge this when you resolve CI.

The CI failure: `ReferenceError: Can't find variable: msg`, seems to be related to your change here,
https://github.com/modeset/teaspoon/pull/480/files#diff-4630ab299de0b26c012329264476c8a5R69 where msg is not defined in the function arguments. I'd say roll that line back since it's unrelated to your other changes, no?

bouk:
ah woops that's a copy paste error, will fix

jejacks0n:
An easy place to start is to unit testing the new setup method.. Check the base teaspoon spec and there'll be some examples. I assume you can call `window.onerror` directly without having to raise an exception, so that would be my first pass. This may not be possible, and if it's not, I'd extract the logic from within the `window.onerror` function, move it to a different method and test that individually.. then just make sure `window.onerror` is the function you expect.

As a follow up, I realized that we could be clobbering an existing window.onerror handler.. you should capture that and call it if it's defined in our own implementation -- pseudo code.

```
existingHandler = window.onerror
window.onerror = =>
   # our logic
   existingHandler?()
```

This also breaks some existing stuff, which is tested.. I have a call to `foo()` in the integration tests which can be found in the spec dummy app assets.. it behaves differently now, and so the output is different.

A few questions now that I thought this through a bit more... when you run jasmine you can disable it's try/catch logic so exceptions are raised -- this makes debugging really nice sometimes. how does this window.onerror handle things now? Should this go only in the console reporters, since that's really the only place we want the behavior?

bouk:
I'm already handling when there is an existing onerror handler.

I think for this to only do its thing when you're in a console report makes sense, but doesn't that mean [just PhantomJS gets this right now](https://github.com/modeset/teaspoon/blob/eae172790dd57fa70b1a9bea3284b0ab79d64b9a/app/assets/javascripts/teaspoon/runner.coffee#L25)? What if someone is using Selenium (which I would assume some people are for CI?)

Also, what do you want me to change about that integration test? Because it's behaving as expected with the new behaviour, do you want the call to `foo()` removed or the expected output to change?

bouk:
@jejacks0n I've added a spec for the onerror handler

What do you want me to do about the integration tests? Should I remove the `foo()` call that makes it fail?

jejacks0n:
To be clear this is why I took the stance I did. When you have an exception it doesn't necessarily break the world -- things can continue running, it's just important to know that it's an issue that you may not have a test for. Anyway, that's the background. =)

If you could remove it and put it into a different test that covers what your error handling accomplishes that would be awesome. So, I'm fine with accepting the adjustment to the behavior, but the tests accurately reflect the change in behavior so we should 1. remove the case that breaks the integration tests (e.g. the call to an undefined function), and add one that ensures that when something like that does happen it fails the build as it does now.

jejacks0n:
To give you some more info, if you check in https://github.com/modeset/teaspoon/blob/c9786f54c487a78e55686f76cf93f199654fea53/teaspoon-jasmine/spec/javascripts/integration/first_integration.coffee you'll see an example of that.. so you'll need to adjust the behavior in each of the different frameworks -- e.g. jasmine, mocha, and qunit.

additionally https://github.com/modeset/teaspoon/blob/c9786f54c487a78e55686f76cf93f199654fea53/teaspoon-jasmine/spec/integration_spec.rb you'll find where we basically configure teaspoon on the fly to get those "real" integration tests into the dummy application.

That's how you'd create a new one basically.. so adjust the existing ones to pass, and then create three new ones (one for each framework) that has a "broken_integration" setup similar to what's already in place.

jejacks0n:
Thanks for taking this on btw.

bouk:
@jejacks0n I've added a broken spec helper test and fixed the other tests, and it has passed CI!

jejacks0n:
I need to exclude the js from hound apparently now that coffeescript -> javascript causes so many issues.

app/assets/javascripts/teaspoon/teaspoon.coffee
```diff
@@ -32,6 +32,16 @@ class @Teaspoon
...
 
 
+  @setupErrorHandler: ->
+    originalOnError = window.onerror
+    window.onerror = (message) ->
+      originalOnError(arguments...) if originalOnError && originalOnError.call
+      return if Teaspoon.started
+      Teaspoon.log JSON.stringify
+        _teaspoon: true
+        type: 'exception'
```
jejacks0n: can you change this to double quotes please?

spec/support/teaspoon_devkit.rb
```diff
@@ -57,6 +57,14 @@ def copy_integration_files(suffix, from, to = "spec")
       end
     end
 
+    def copy_broken_helper(suffix)
+      test_directory = expand_path("#{suffix}/javascripts")
+      FileUtils::mkdir_p(test_directory)
+      broken_helper = File.expand_path('../../../spec/support/broken_spec_helper.coffee', __FILE__)
```
jejacks0n: same with the double quotes.

bouk:
@jejacks0n I've updated the hound config to exclude the generated files

teaspoon-qunit/spec/broken_helper_spec.rb
```diff
@@ -0,0 +1,22 @@
+require_relative "./spec_helper"
+
+feature "Broken helper", shell: true do
+  let(:version) do
+    Teaspoon::Framework.fetch(:qunit).versions.last
+  end
+
+  before do
+    teaspoon_test_app(<<-GEMFILE)
+      gem 'teaspoon', path: '#{Teaspoon::DEV_PATH}'
```
jejacks0n: double quotes, and I think you might not need the interpolation here.

jejacks0n:
Cool, looks good.. will merge after CI is done.

Thanks again for all your work here, much appreciated.

teaspoon-qunit/spec/broken_helper_spec.rb
bouk: I just copied this verbatim from console_spec.rb

bouk:
CI is done, and no problem! I'd much appreciate it if you cut a release 😄

jejacks0n:
I'm not planning on releasing today because I'm pairing on some other things, but hit me up in a few days if I've forgotten.

bouk:
@jejacks0n just checking in, don't think there's been a release yet

winniehell:
@jejacks0n Do I get it right that there is still no release containing this? Judging from the changelog, 1.1.5 in March was the last release.

PR #485, opened 2016-06-01
Don't log error event to console

bouk:
Just ran into an interesting issue: because of same origin policies, javascript errors that get caught in `onerror` won't have an error message if they happen on a JavaScript file on a different domain.

I have modified the phantomjs onerror handler to log as "exception" instead of "error" when Teaspoon hasn't started yet, so it has the same effect as the `onerror` handler, just with more info available


bouk:
for review @jejacks0n

PR #498, opened 2016-07-27
Suite controller now returns correct status codes, and updated js to …

### LINKED ISSUE #493: Hook produces ReferenceError: url is not defined
Hey!

I try to create a hook like so:

(*teaspoon_env.rb*, in `config.suite do |suite|`)
```
suite.hook :create_document do
      require 'factory_girl_rails'

      MyDocument.destroy_all

      FactoryGirl.definition_file_paths = ['./spec/factories']
      FactoryGirl.find_definitions
      FactoryGirl.create(:annotation_document)
    end
```

and call it in my *my_test_spec.coffee* with

```
before ->
  Teaspoon.hook('create_document')
```

This leads me to the following error in both, browser and CLI:

>  "before all" hook
> ReferenceError: url is not defined (http://localhost:3000/assets/teaspoon-mocha.self-b15d919...

What am I doing wrong? Thanks a lot! : )

ZackMattor:
…handle hook status codes.

Fixes #493 

 - [x] Fix `SuitesController` so it doesn't always return a 500 status
 - [x] `SuitesController` should return a 200 on success, and a 404 when the hook can't be found.
 - [x] All hook xhr javascript should throw the error returned by the server

### DIFF
--- a/app/assets/javascripts/teaspoon/hook.coffee
+++ b/app/assets/javascripts/teaspoon/hook.coffee
@@ -19,4 +19,4 @@ Teaspoon.hook = (name, payload = {}) ->
 
   xhrRequest "#{Teaspoon.suites.active}/#{name}", payload, ->
     return unless xhr.readyState == 4
-    throw("Unable to call hook \"#{url}\".") unless xhr.status == 200
+    throw JSON.parse(xhr.response).err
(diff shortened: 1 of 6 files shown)


jejacks0n:
:+1:

ZackMattor:
There should probably be some tests around this... But that's all I can do today.

ZackMattor:
@jejacks0n I think we should handle 500 errors better. Because currently it tries to parse them as json and throws a json parse error. I'll see if I can come up with a good way to handle that. Probably just 

```
 if(xhr.status === 500) throw "Server returned a 500, check your hook's code."
```

PR #499, opened 2016-08-18
Make boot partial template consistent with default

mhaylock:
The default `_boot.html.erb` partial had an additional `allow_non_precompiled: true` option added in 1e3a22311fc505caf70e20f752a59e19a2eab9d2, but the template for the generator was not update at the same time.

When I ran `rails generate teaspoon:install --partials` and attempted to use the generated boot partial without modifications I found teaspoon was failing to load. Updating the template to match the default fixes this.


jejacks0n:
Thanks!

PR #511, opened 2017-01-24
Adding support for multiple suites through options

tsov:
Currently it seems _impossible_ to run selected suites through the command line, other than using the files options. The problem with using the files option, if you have __a lot of test files__, the query string can get so big that the following error gets thrown: `Invalid request: HTTP element QUERY_STRING is longer than the (1024 * 10) allowed length.`

There might be a better way to allow running multiple suites than what I suggest, but this is definitely a missing feature 😕


mathieujobin:
I rebased this change into 25d3ce0122a64aaf161f8d0656450d64e4e5fce8

all tests seems to pass, if someone else thinks there is value here, I can merge this

mathieujobin (approved):

mathieujobin:
released in v1.4.0
https://rubygems.org/gems/teaspoon/versions/1.4.0

PR #520, opened 2017-05-24
Add support for precompiled assets manifest in Sprockets >= 3.0

mankalas:
Sprockets 3.0 has changed the way assets can be accessed if the fallback
asset pipeline is disabled (Rails config option `config.assets.compile`
is set to false).

In such case scenario, extracting asset information based on
Sprockets environment doesn't work, since `Rails.application.assets`
returns nil.

Fortunately, similar information about assets can be obtained from asset
manifest, accessible from `Rails.application.assets_manifest`.

This change makes teaspoon fall back to using the assets manifest if
regular asset environment is not available. It seems to work for all the
use cases that I tested.

### DIFF
--- a/lib/teaspoon/suite.rb
+++ b/lib/teaspoon/suite.rb
@@ -21,6 +21,7 @@ module Teaspoon
       @name = (@options[:suite] || :default).to_s
       @config = suite_configuration
       @env = Rails.application.assets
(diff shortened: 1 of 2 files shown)


mathieujobin:
please update your branch with latest master

mathieujobin:
feel free to reopen when you are ready to resume work on it.

Issue #405, opened 2015-09-08
Missing CoffeeScript dependency?


benpickles:
# Problem

In Rails `rake assets:precompile` blows up with `LoadError: cannot load such file -- coffee_script` in development/test environments.

# Steps to reproduce

1. Generate a new Rails app.
2. Run `rake assets:precompile` - it works!
3. Add the Teaspoon gem to the Gemfile: `gem 'teaspoon', groups: [:development, :test]`.
4. Run `rake assets:precompile` - it does not work :frowning: 

This is with Rails 4.2.4 and Teaspoon 1.0.2.

```
$ rake assets:precompile
I, [2015-09-08T09:59:07.434046 #84070]  INFO -- : Writing /.../public/assets/teaspoon-5a63f1400783e117b9e55161f53688b6530615d285f32af0abde8f252706d140.css
I, [2015-09-08T09:59:07.435940 #84070]  INFO -- : Writing /.../public/assets/teaspoon-filterer-65c5f62e0af566bf9d24f413533144277cb7d2fa9461e8307332fede303af366.js
I, [2015-09-08T09:59:07.438136 #84070]  INFO -- : Writing /.../public/assets/application-e80e8f2318043e8af94dddc2adad5a4f09739a8ebb323b3ab31cd71d45fd9113.css
I, [2015-09-08T09:59:07.440054 #84070]  INFO -- : Writing /.../public/assets/support/bind-poly-c89d2f5f622318bb7bd38eb36472131e70bdb95e45431cd9397a192fa3fd24f7.js
I, [2015-09-08T09:59:07.442498 #84070]  INFO -- : Writing /.../public/assets/support/phantomjs-shims-f6867d3adaf9c1a659d0ca8e170f147e97bbde37736890cf05d84e7b5020fed4.js
I, [2015-09-08T09:59:07.447751 #84070]  INFO -- : Writing /.../public/assets/support/sinon-3dade19b521c697425009cf3586fe148d7ef3cae04beb8561ce9efd984a50242.js
rake aborted!
LoadError: cannot load such file -- coffee_script
/.../activesupport-4.2.4/lib/active_support/dependencies.rb:274:in `require'
/.../activesupport-4.2.4/lib/active_support/dependencies.rb:274:in `block in require'
/.../activesupport-4.2.4/lib/active_support/dependencies.rb:240:in `load_dependency'
/.../activesupport-4.2.4/lib/active_support/dependencies.rb:274:in `require'
[...]

PR #528, opened 2017-09-27
Add coffee-rails dependency

md5:
Scaffolded Rails projects include `coffee-rails` unless they use `--skip-coffee`, but not all applications include it.

This PR makes the dependency explicit.

See #405

### DIFF
--- a/teaspoon.gemspec
+++ b/teaspoon.gemspec
@@ -18,5 +18,6 @@ Gem::Specification.new do |s|
   s.files       = Dir["{app,lib,bin}/**/*"] + ["MIT.LICENSE", "README.md", "CHANGELOG.md"]
   s.executables = ["teaspoon"]
 
+  s.add_dependency "coffee-rails", [">= 3.2.2", "< 5"]
   s.add_dependency "railties", [">= 3.2.5", "< 6"]
 end


md5:
Looks like the Travis failures are not specific to this branch. I see the same failures for other recent PRs.

BrianHawley:
Please don't. This will add the coffee-rails dependency to projects that don't have or want it. It's better to make a change to the README or wiki to mention this issue.

md5:
@BrianHawley That doesn't make any sense. The fact of the matter is that `teaspoon` requires `coffee-rails`. You simply can't use it in a project that does not have `coffee-rails` installed somehow.

BrianHawley:
If you don't use the `--coffee` option and don't write your specs in CoffeeScript then you don't need coffee-rails. Teaspoon probably could use coffee-rails as a development dependency, to convert its own source from CoffeeScript to js, and it needs it for its tests. However, it doesn't need it as a packaged gem because all of the js is preconverted. If you as a *user* of Teaspoon don't use CoffeeScript, it should still work.

md5:
@BrianHawley Where is this preconverted JS you speak of? I only see Coffeescript files here: https://github.com/jejacks0n/teaspoon/tree/master/app/assets/javascripts/teaspoon

Are you not using asset precompilation? The problem is that if you run `assets:precompile`, the Sprockets pipeline attempts to compile the `*.coffee` files from this gem and fails if it can't `require 'coffee_script'`.

md5:
Ah, I see them now: https://github.com/jejacks0n/teaspoon/tree/master/teaspoon-jasmine/lib/teaspoon/jasmine/assets

md5:
Actually, I see a bunch of `*.coffee` files there as well...

BrianHawley:
It looks like we only run the tests in CI mode, which doesn't use the browser UI, with its coffee files. Weird that we've been using this for a year without coffee-rails and it's never caused a problem for us. We should consider adding coffee-rails as a development dependency ourselves.

BrianHawley:
Nope, I was wrong, the coffee files are precompiled by teaspoon-devkit.rb in Teaspoon's development mode. I'm going to check the generated files in the installed's gem's directory to be sure.

md5:
Looks like the `teaspoon/*.js` here may be the source of this problem: https://github.com/jejacks0n/teaspoon/blob/0335e79b385c2a74d2d963417b8857368a6cc928/lib/teaspoon/configuration.rb#L24

It's unclear to me which `*.js` files this is attempting to precompile, but this is causing Sprockets to try to process files like `app/assets/javascripts/teaspoon/error.coffee`. As far as I can tell (thanks to you pointing it out @BrianHawley), these Coffeescript files should already be precompiled into individual driver files (like `teaspoon-jasmine/lib/teaspoon/jasmine/assets/teaspoon-jasmine2.js`).

BrianHawley:
Looks like the coffee files exist in the "app/assets/javascripts/teaspoon" directory when the gem is installed, so the js files aren't precompiled. I should look into this, because not having coffee-rails as a runtime dependency would be considered an advantage to people who don't want it.

md5:
@BrianHawley I'm not a fan myself

BrianHawley:
I checked with our devs and they use the browser UI, so that means that the coffee files are compiled somehow without using coffee-rails. It's possible that the old sprockets version we use has coffee support built-in and doesn't need coffee-rails.

BrianHawley:
Yup, that was it. Our version of sprockets uses tilt to handle the coffee files, no coffee-rails needed. It looks like Teaspoon's development dependencies are handled by the teaspoon-devkit gemspec, which has coffee-rails as a dependency already.

That means that the best solution for your problem is to document that if you're using sprockets 3+, you need to add coffee-rails as a dependency. I think the Installation section of the README is the best place to put it.

md5:
@BrianHawley I disagree that that's the best solution since Sprockets 4 is the current version and earlier versions are not maintained.

md5:
After applying the following diff, I was able to run my test from https://github.com/jejacks0n/teaspoon/issues/405#issuecomment-323471804 (adjusting to point at my local copy of `teaspoon`):

```diff
commit c8320b0214ef4d7521aac5acecd5a0d3b07faf9a
Author: Mike Dillon <mike.dillon@disney.com>
Date:   Fri Dec 29 12:41:42 2017 -0800

    Remove wildcards for teaspoon/*.js and support/*.js from precompile paths

diff --git a/lib/teaspoon/configuration.rb b/lib/teaspoon/configuration.rb
index 00ee7b7..f9b48d8 100644
--- a/lib/teaspoon/configuration.rb
+++ b/lib/teaspoon/configuration.rb
@@ -21,7 +21,7 @@ module Teaspoon
     @@asset_paths    = ["spec/javascripts", "spec/javascripts/stylesheets",
                         "test/javascripts", "test/javascripts/stylesheets"]
     @@fixture_paths  = ["spec/javascripts/fixtures", "test/javascripts/fixtures"]
-    @@asset_manifest = ["teaspoon.css", "teaspoon-filterer.js", "teaspoon/*.js", "support/*.js"]
+    @@asset_manifest = ["teaspoon.css", "teaspoon-filterer.js"]
 
     # console runner specific
 
```

md5:
That diff is probably too aggressive. It looks like it would be OK to leave `support/*.js` (or just explicitly list the three `support/*.js` files).

md5:
Closing this in favor of #536

BrianHawley:
Sprockets 4 is in beta, not the current version; Sprockets 3.5.2 is the current version. Sprockets 2.x is still in wide use, even though it's not currently maintained, so it's not a good idea to drop it when you don't really have to. In any case, there are several different CoffeeScript processors in common use, so there's no reason for Teaspoon to take sides.

PR #519, opened 2017-05-05
Allow client options to be passed into Jasmine

jsilvestri:
This addresses https://github.com/jejacks0n/teaspoon/issues/518. Example usage in teaspoon_env.rb:

```
client = Selenium::WebDriver::Remote::Http::Default.new
client.timeout = 180 # 3 mins instead of the default 60 seconds
config.driver_options = {client_driver: :firefox, client_driver_opts: {http_client: client}}
```

PR #530, opened 2017-11-14
Jasmine client options

tilsammans:
Continues #519 

Thanks to everyone involved!

### DIFF
--- a/lib/teaspoon/driver/selenium.rb
+++ b/lib/teaspoon/driver/selenium.rb
@@ -23,7 +23,7 @@ module Teaspoon
       end
 
       def run_specs(runner, url)
-        driver = ::Selenium::WebDriver.for(driver_options[:client_driver], driver_options[:client_driver_opts] || {})
+        driver = ::Selenium::WebDriver.for(driver_options[:client_driver], @options[:client_driver_opts]&.dup || {})
         driver.navigate.to(url)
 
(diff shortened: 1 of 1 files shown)


lib/teaspoon/driver/selenium.rb
```diff
@@ -23,7 +23,7 @@ def initialize(options = nil)
       end
 
       def run_specs(runner, url)
-        driver = ::Selenium::WebDriver.for(driver_options[:client_driver])
+        driver = ::Selenium::WebDriver.for(driver_options[:client_driver], @options[:client_driver_opts]&.dup || {})
```
zzak: Doesn't this also make a dependency on whatever Ruby version introduced this new syntax? I think 2.3?

jsilvestri:
`options[:client_driver_opts].symbolize_keys` works as well and might address the syntax concerns here: https://github.com/jsilvestri/teaspoon/commit/7ec4b6d6e599cfe1a446ceea84a2e674550b0ca6

zzak:
@jsilvestri I believe `Hash#symbolize_keys` is only available in ActiveSupport, but not sure that is a concern here.

jsilvestri:
@zzak correct.  Agreed it may not be ideal to depend on Active Support.

mathieujobin:
looks like #537 properly fix this issue, please reopen if you disagree

PR #535, opened 2017-12-13
Modified to work with Sprockets 4 beta 6

maxhudson:
I was able to get Sprockets 4 beta 6 + es6 javascript files working with teaspoon by making this modification and setting `config.expand_assets = false`.

Teaspoon is using all of these and throwing errors because of them:

Has a deprecated dependency as far as I can tell (.pathname): http://www.rubydoc.info/github/sstephenson/sprockets/master/Sprockets%2FAsset:logical_path
Deprecated: http://www.rubydoc.info/github/sstephenson/sprockets/Sprockets%2FAsset%3Apathname
Deprecated: http://www.rubydoc.info/github/sstephenson/sprockets/Sprockets%2FAsset%3Ato_a

The way I modified the `suite.rb` file probably isn't ideal, but it does work for me. It would probably need to be modified if this were to merge.

### DIFF
--- a/lib/teaspoon/suite.rb
+++ b/lib/teaspoon/suite.rb
@@ -64,8 +64,8 @@ module Teaspoon
       params = []
       params << "body=1" if config.expand_assets
       params << "instrument=1" if instrument_file?(asset.filename)
-      puts asset.filename
-      url = asset.filename.gsub(/^(.*?)\javascripts/, '')
+      puts asset.filename.gsub(/^(.*?)\javascripts\//, '')
+      url = asset.filename.gsub(/^(.*?)\javascripts\//, '')
       url += "?#{params.join("&")}" if params.any?
       url
     end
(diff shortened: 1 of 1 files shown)


jejacks0n:
I don't think this is a complete solution.. you can have your javascripts in any path, simply by adding paths to the asset paths configuration.

Because of that, I don't see how this is something that could be merged.

So is there any news about what the deprecations are replaced with?

jejacks0n:
I really appreciate your efforts on this -- this is by no means a slight, I just think asking a maintainer of sprockets might be a better solution -- see if there's an alternative, explain the need for it, etc.

maxhudson:
I understand. I figured I'd make a pull request in-case you had a real solution.

From what I could tell, `.filename` is supposed to be used instead of pathname. `.logical_path` is not actually deprecated, but an error is thrown when it is used saying that `.pathname` is not a method, so it seems like `.logical_path` should be deprecated or modified in sprockets. That specific point seems like something worth mentioning to @sstephenson @rails.

As far as `.to_a`, I'm not sure, but the default configuration of teaspoon seems to rely upon it because the default `config.expand_assets` is set to true and results in that method being called.

mathieujobin:
I am guessing this is outdated and could be closed?

mathieujobin:
looks like this can be closed

PR #543, opened 2018-05-31
Support Jasmine 2.9.1 which includes async/await support

sblackstone:
Add support for Jasmine 2.9.1 which appears to be the last version before 3.0 which had breaking changes.


josuemontano:
@sblackstone any news on this?

PR #544, opened 2018-07-05
Update changelog from last release

aharpervc:
- after https://github.com/jejacks0n/teaspoon/pull/536 & others, the minimum version needs to be bumped for release
- add changelog entry
- closes https://github.com/jejacks0n/teaspoon/issues/494

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,18 @@
 ### Unreleased
 
-* Adds support for Rails 6.
-* Removes build support for Rails 3 & 4.
-* Removes build support for Ruby < 2.5.
-* Removes support for capybara webkit.
+### 1.2.0
+
+#### Enhancements
+
+* Add Rails 6 (beta1) support (#556)
+* Allow passing selenium options (#537)
 
 #### Bug Fixes
+
+* Removes Rails 3 & 4 support (#560)
+* Removes support for capybara webkit (#560)
+* Fixes gem coffeescript assets being discovered by sprockets (#405)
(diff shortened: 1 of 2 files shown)


aharpervc:
@jejacks0n this is the same CI failure as the previous commit on master, so it doesn't appear to have been introduced by this branch. Also, I can't reproduce locally (all tests pass both on master and this branch).

Any idea what's going on?

Previous failure: https://travis-ci.org/jejacks0n/teaspoon/builds/399620721

twalpole:
@aharpervc The failures are because the tests are using Capybara 3+ which defaults to using `puma` as the server.  You either need to include `puma` in the project, specify `Capybara.server = :webrick` in the configuration or lock Capybara to < 3.  If you update the project gems locally you should see the error locally too.

aharpervc:
I suppose there's a question if this should be a major version change, since old Rails support was dropped. I'm neutral on this question, I just want the latest code via rubygems :)

twalpole:
SemVer technically allows a package to change it's minimum requirements on a minor version - as long as code written to the packages API doesn't break - https://semver.org/#what-should-i-do-if-i-update-my-own-dependencies-without-changing-the-public-api

jejacks0n:
I think we should to a min version update at least, no?

aharpervc:
I switched it to `1.2.0`, although does that mean that mocha, qunit, and jasmine also need minor bumps? Probably not since that relationship is very flexible (`>= 1.0.0`)?

PR #551, opened 2018-12-07
Trying to fix the build + Selenium options

pablobm:
*Don't merge.* This is the same as https://github.com/jejacks0n/teaspoon/pull/537, but with an additional commit to attempt fix the breaking build.


pablobm:
Took this to https://github.com/jejacks0n/teaspoon/pull/552

PR #561, opened 2019-04-16
Move inline JS to rendered files to support CSP

worldwise001:
As it is currently set up, if you're running Teaspoon in apps that have strong CSP settings, then all the inline JS pieces will not load causing people to be confused as to why individual tests are not running.

This patch fixes that by moving the files into their own JS files. The trick though was that we still need the suite information for the JS, so what we do here is that we add another route into the controller to custom render the JS based on the suite name (which we also use for the show view anyway).

### DIFF
--- a/app/controllers/teaspoon/suite_controller.rb
+++ b/app/controllers/teaspoon/suite_controller.rb
@@ -16,6 +16,12 @@ class Teaspoon::SuiteController < ActionController::Base
     @suite = Teaspoon::Suite.new(params)
   end
 
+  def js
+    jstype = params.extract!(:jstype).fetch(:jstype, 'vanilla')
+    @suite = Teaspoon::Suite.new(params)
+    render "teaspoon/suite/#{jstype}", :content_type => 'application/javascript'
+  end
+
   def hook
     hooks = Teaspoon::Suite.new(params).hooks[params[:hook].to_s]
 
--- a/app/views/teaspoon/suite/_boot.html.erb
+++ b/app/views/teaspoon/suite/_boot.html.erb
@@ -1,4 +1,2 @@
 <%= javascript_include_tag *@suite.spec_assets, debug: @suite.config.expand_assets, allow_non_precompiled: true %>
-<script type="text/javascript">
-  Teaspoon.onWindowLoad(Teaspoon.execute);
-</script>
+<script src="<%= "#{Teaspoon.configuration.mount_at}/js/#{@suite.name}.vanilla.js" %>"></script>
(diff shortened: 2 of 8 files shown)


worldwise001:
cc @jejacks0n

jejacks0n:
This is awesome, thanks @worldwise001! I'm in the process of getting this project up to speed and more current, and I'd like to work with you on getting this in there.

My plan is specifically to focus around CSP settings, webpack(er), and reducing the asset pipeline dependency in that process.

Maybe you can help -- I'm seeing calls to `require`, how is this being provided? is it already in the project?

worldwise001:
Yes, it should all be already in the project. I just moved some files around so that they are served via the controller instead of being inline JS.

jejacks0n:
Seems to be! answered my own question. haha.

Ok, so one thing to note here is that right now not all specs / features are being run on CI since I'm still working on getting things polished up. When I get CI back to a place I'm comfortable with can I hit you up to merge master and we can check in at that point?

worldwise001:
(fwiw I tested this using my own project so I can verify that this causes the CSP errors to go away... I haven't really tested require.js since I don't use that in my project).

worldwise001:
Sure! Let me know when you're likely going to be able to get CI in a nice place and I can retest.

jejacks0n:
Thanks so much @worldwise001 -- I appreciate your time and will reach out in the next couple days as time permits. =)

pamelahowell:
Assume sauce! Kudos to you both, and sounds like the start of a lovely collaboration.

jejacks0n:
@worldwise001 -- The test suite is running fully and coverage reports are back in there now. There's a flakey test about code coverage, and I haven't resolved that because it's pretty complex -- and seems like a caching/random spec order issue that manifests somewhere inside sprockets.

You're welcome to rebase master now, and just keep an eye out for that one test. You can just re-run the failing build(s) until it passes -- sorry about that!

Thanks!

worldwise001:
Fantastic! I'll do that today! (sorry been swamped with other unrelated work).

worldwise001:
Hrm it looks like something is rewriting all the routes in the test to add /relative in front of it. I imagine this is to test something, but seems to break all the teaspoon routes. Is this intended? I think this is only supposed to kick in for assets?

worldwise001:
(Any hints on where to look? I tried adding binding.pry to the erbs and those seem to be generating properly, but I couldn't figure out where in the stack frame the rewrite was occurring)

worldwise001:
never mind, fixed it. `javascript_include_tag` assumed that it was an asset and so added the relative path accordingly.

jejacks0n:
Ignore hound.. sorry.

jejacks0n:
So, do those failures make any sense to you locally?

jejacks0n:
I changed up the rails versions a bit.

worldwise001:
The failures are a little baffling to me, and I'm trying to repro it locally. This might take the weekend for me to figure out.

mathieujobin:
several tests is failing with this changed https://github.com/jejacks0n/teaspoon/actions/runs/657166943
rebased into e85199b

PR #562, opened 2019-06-01
Fix NoMethodError: undefined method `root' for Teaspoon:Module when running the console

CyborgMaster:
Commit 5b912da349a11d12ccdf1705a1aeb79e4f1e3a57 uses the utility method `Teaspoon.root` to find paths.  The console also needs access to this method or guard-teaspoon dies on boot.

### DIFF
--- a/lib/teaspoon/console.rb
+++ b/lib/teaspoon/console.rb
@@ -1,4 +1,5 @@
 require "teaspoon/environment"
+require "teaspoon/utility"
 
 module Teaspoon
   class Console


shelmire:
@jejacks0n I've inherited CyborgMaster's code and I'd like to keep using Teaspoon with guard. Any way we could get this merged?

mathieujobin:
@jejacks0n how about we create a `teaspoon` team, and move the repo to it.
invite a few folks willing to cut a release and do code reviews.

I can do this at the very minimum.

mathieujobin:
@jejacks0n ping ! lets create a team and move your repo to it 
the only reason I am against a fork, is because we need to rename the gem on rubygems.org which is annoying to the few users we have

PR #568, opened 2019-08-19
Update to support Rails 6

jtapia:
### Description

- Fix issue `Could not find Selenium Webdriver. Install the selenium-webdriver gem.`
- Fix issue `sh: 0: Can't open /etc/init.d/xvfb - The command "sh -e /etc/init.d/xvfb start" failed and exited with 127 during .` on CI
- Add script to install `phantomjs` on CI

### DIFF
--- a/teaspoon-devkit.gemspec
+++ b/teaspoon-devkit.gemspec
@@ -29,6 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency "aruba"
   s.add_dependency "appraisal"
   s.add_dependency "simplecov"
+  s.add_dependency "selenium-webdriver"
 
   # for local bundle installs
   s.add_dependency "jbuilder"
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ env:
 before_script:
   - sh -e /etc/init.d/xvfb start
   - npm install -g istanbul
+  - npm install -g phantomjs
   - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > ./cc-test-reporter
   - chmod +x ./cc-test-reporter
   - ./cc-test-reporter before-build
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 sudo: false
 cache: bundler
 rvm:
--- a/Appraisals
+++ b/Appraisals
@@ -3,5 +3,5 @@ appraise "rails_5" do
 end
 
 appraise "rails_6" do
-  gem "rails", "6.0.0.beta3"
+  gem "rails", "6.0.0"
 end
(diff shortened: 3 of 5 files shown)


jejacks0n:
If you've confirmed it works, I'm happy to merge.

jejacks0n:
Will circle back and add rails 6 to the test matrix.

jejacks0n:
oh, actually, you can go ahead and update

https://github.com/jejacks0n/teaspoon/blob/master/Appraisals

as well. That way we make sure what we claim we support, is tested.

teaspoon.gemspec
```diff
@@ -20,4 +20,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = "~> 2.4"
   s.add_dependency "railties", ">= 3.2.5"
+  s.add_dependency "capybara"
+  s.add_dependency "selenium-webdriver"
+  s.add_dependency "webdrivers"
```
jejacks0n: these are already over here: https://github.com/jejacks0n/teaspoon/blob/master/teaspoon-devkit.gemspec

jejacks0n: and they shouldn't be dependencies of teaspoon, because you might not use them, unless they are now? haha

teaspoon-devkit.gemspec
```diff
@@ -29,6 +29,7 @@ Gem::Specification.new do |s|
   s.add_dependency "aruba"
   s.add_dependency "appraisal"
   s.add_dependency "simplecov"
+  s.add_dependency "selenium-webdriver"
```
jejacks0n: This still isn't a requirement -- but we could maybe update the readme? phantomjs is still an option for using as a driver, and it might be annoying to require `selenium-webdriver` if that's what you're using. Open to opinions.

jtapia: yes, I'm just testing, but I think the issue is only on the CI, if I got anything, I'll let you know

jejacks0n: ah, ok. haha. try taking out the line it's failing on -- maybe we don't need to do that anymore?

jtapia: now is green, but I will check about the broken test, let me know what do you think

mileslane:
Will this work well with the selenium webdriver / chromedriver?

mauriciozaffari:
@jejacks0n are you planning merge this and release? I'm already using @jtapia fork, but would be nice to have an official release.

meowsus:
@jtapia Great work! We've been using this for a few weeks while upgrading Workarea to be Rails 6 compatible and everything is running smoothly.

@jejacks0n a merge and release would be most appreciated!

jejacks0n:
Thanks for your work here! Please confirm by pointing your gem at master, and I've put it on my calendar to give it a week before I release a new version.

meowsus:
@jejacks0n Just a friendly reminder that we're still waiting for a release :*

PR #579, opened 2020-03-11
Fix coverage output path

ukdave:
The output directory argument wasn't being passed to istanbul correctly causing the coverage report to always get generated in `./coverage/` instead of what had been configured in teaspoon.

I've updated the `Open3.capture2e` call in `lib/teaspoon/coverage.rb` to pass the argument names and values separately which fixes the issue.


mathieujobin:
please merge good PRs like this one, so we can move forward.

I need teaspoon to work with latest selenium

PR #582, opened 2020-08-20
Delay using Rails until loaded

eric-hemasystems:
When using ActionController, instrumentation should be used to wait until the component being used is fully loaded. Without this, it triggers the loading early through the autoloader which can lead to other problems.

For more background see rails/rails#36546.


mathieujobin (approved):
seems to make sense

@jejacks0n opinion?

jejacks0n:
I’m am way overloaded with life things, so support whatever choices are best for everyone. Thank you @mathieujobin. 👍

mathieujobin:
I vote we first enable tests to run on Github Actions #588 
Then ask @eric-hemasystems to rebase this PR
and merge it does not introduce problem. but it appears to be a good change to me

PR #588, opened 2021-03-09
Add github actions build.yml file

mathieujobin:
Attempt at making test work on Github Actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,9 +45,9 @@ jobs:
         ruby-version: ${{ matrix.ruby }}
         bundler-cache: true
     - name: Download istanbul
-      run: npm install -g istanbul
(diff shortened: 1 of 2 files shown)


mathieujobin:
jruby has the least failure as far as I can tell.

there are different spec failures depending on which ruby is loaded

https://github.com/mathieujobin/teaspoon/runs/2062185057?check_suite_focus=true

mathieujobin:
last run

https://github.com/mathieujobin/teaspoon/actions/runs/634347827

mathieujobin:
@jejacks0n can we get this merged so we get the ball rolling on running tests on github actions.
then we can ask all PR authors to rebase and see if their change cause any problems or fix issues.

test no longer run on travis it appears

PR #595, opened 2023-02-20
passing --filter="...." correctly escapes the spaces into "%20" and not "+"

thebravoman:
PR created against 1.2.2

Running teaspoon as

```bash
bundle exec teaspoon -s default --filter="ScenePointers are attached"
```

When using CGI.escape the string "ScenePointers are attached" becomes "ScenePointesr+are+attached" This is then passed as a "grep=ScenePointers+are+attached" into the url

In the JS that is running the specs in the browser the value of the grep is used to filter. We move through all the specs and encode their full name with encodeURIComponent

The result is

```javascript
encodeURIComponent("ScenePointers are added")
'ScenePointers%20are%20added'
```

This means that on the ruby side we were escaping the space " " in the urls as '+' and on the JS side we were escaping them as "%20". One of these had to change and it seems changing it on the ruby side is the right one.

There is no right way to escape. We can use 'addressable' and URI and CGI and ERB. ERB seems to be the most straightforward for this case, it also does not bring a lot of dependencies. Just one - 'cgi'.

In the same time there seems to be no method in URI or CGI or 'addressable' gem that would address this

### DIFF
--- a/lib/teaspoon/console.rb
+++ b/lib/teaspoon/console.rb
@@ -1,5 +1,6 @@
 require "teaspoon/environment"
 require "teaspoon/utility"
+require "erb"
 
 module Teaspoon
   class Console
@@ -108,7 +109,10 @@ module Teaspoon
 
       def filter(suite)
         parts = []
-        parts << "grep=#{CGI.escape(options[:filter])}" if options[:filter].present?
+        # Using ERB::Util.url_encode instead of CGI.escape as CGI.escape will convert
+        # the space ' ' into a '+' and not into "%20". The JS on the front end expects
+        # %20
+        parts << "grep=#{ERB::Util.url_encode(options[:filter])}" if options[:filter].present?
         (@suites[suite] || options[:files] || []).flatten.each { |file| parts << "file[]=#{CGI.escape(file)}" }
         "#{parts.join('&')}" if parts.present?
       end
--- a/teaspoon.gemspec
+++ b/teaspoon.gemspec
@@ -20,5 +20,6 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = ">= 2.4"
   s.add_dependency "railties", ">= 3.2.5"
+  s.add_dependency "erb"
   s.add_development_dependency "simplecov", "< 0.18"
 end


mathieujobin:
Did you check why the CI isn't happy ?

thebravoman:
I did. It was too much and I could not directly see the connection between the change the the fails.

Let me know what you think about the PR. If you think it is in the right direction I could invest some time to understand why the specs fail.

teaspoon.gemspec
```diff
@@ -18,7 +18,8 @@ Gem::Specification.new do |s|
   s.files       = Dir["{app,lib,bin}/**/*"] + ["MIT.LICENSE", "README.md", "CHANGELOG.md"]
   s.executables = ["teaspoon"]
 
-  s.required_ruby_version = ">= 2.5"
-  s.add_dependency "railties", ">= 4.2.11"
+  s.required_ruby_version = ">= 2.4"
+  s.add_dependency "railties", ">= 3.2.5"
```
mathieujobin, line 22: Why allowing ruby 2.4 and rails 3.2 ?

lib/teaspoon/console.rb
```diff
@@ -108,7 +109,10 @@ def url_for(suite, console = true)
 
       def filter(suite)
         parts = []
-        parts << "grep=#{CGI.escape(options[:filter])}" if options[:filter].present?
+        # Using ERB::Util.url_encode instead of CGI.escape as CGI.escape will convert
+        # the space ' ' into a '+' and not into "%20". The JS on the front end expects
+        # %20
+        parts << "grep=#{ERB::Util.url_encode(options[:filter])}" if options[:filter].present?
```
mathieujobin, line 115: that seems to make sense

mathieujobin:
you're right failures, seems quite unrelated

please have a look at https://github.com/jejacks0n/teaspoon/pull/597

and please rebase once merged

adding a test to reproduce the problem you are fixing would be great

thebravoman:
Thanks for the feedback. Now as this makes sense I will think of a spec, rebase and clean up the teaspoon.gemspec

mathieujobin:
please update your branch with latest master

lib/teaspoon/console.rb
```diff
@@ -1,5 +1,6 @@
 require "teaspoon/environment"
```
captain-hoity-toity, line 1: Overall, the code is well written and follows good practices. However, there are a few points that could be improved:

1. **Error Handling**: There are several instances where exceptions are caught and simply logged or re-raised. It would be better to handle these exceptions in a more user-friendly way, or avoid raising them in the first place by using proper checks and validations.

2. **Method Length**: The `execute_without_handling` method is quite long and does a lot of things. Consider breaking it down into smaller, more manageable methods. This would improve readability and maintainability of the code.

3. **Use of Instance Variables**: There are several instance variables being used throughout the class. Consider encapsulating them into methods, or reducing their scope if possible. This would make the code easier to understand and debug.

4. **Use of `rescue` in method definitions**: It's generally considered bad practice to use `rescue` in method definitions as it can lead to unexpected behavior and makes the code harder to follow. Consider moving the error handling to a separate method.

5. **Code Comments**: While the code is generally well-commented, there are some areas where the purpose of the code is not immediately clear. Adding more detailed comments in these areas would be helpful.

6. **Naming**: Some of the method names are not very descriptive. Consider renaming methods to more accurately describe what they do. For example, `filter` could be renamed to `filter_url_params`. 

Here is an example of how to refactor the `execute_without_handling` method:

```ruby
def execute_without_handling(execute_options = {})
  @execute_options = execute_options
  resolve(options[:files])

  run_all_suites
end

def run_all_suites
  suites.inject(0) do |failures, suite|
    export(suite) if options.include?(:export)
    failures += run_suite(suite)
    log("") # empty line for space
    failures
  end
end

def run_suite(suite)
  raise Teaspoon::UnknownSuite.new(name: suite) unless Teaspoon.configuration.suite_configs[suite.to_s]

  log("Teaspoon running #{suite} suite at #{base_url_for(suite)}")
  runner = Teaspoon::Runner.new(suite)
  driver.run_specs(runner, url_for(suite))
  check_for_failures(runner)
  runner.failure_count
end

def check_for_failures(runner)
  raise Teaspoon::Failure if Teaspoon.configuration.fail_fast && runner.failure_count > 0
end
```

This way, each method has a single responsibility and the code is easier to follow.

mathieujobin (approved):

mathieujobin (changes requested):
not sure why, but existing test fails.
and I thought you had added a new test for this fix. but I am not seeing it.

Thanks for your contribution.

hopefully it helps keeping Teaspoon alive for Rails 7.x and 8.x future.

PR #602, opened 2024-04-04
Use `exit!` instead of `exit`

ghiculescu:
We found that using `exit` here would conflict with the `debug` gem on Ruby 3.2. It would hang and never exit when tests completed, similar to this issue: https://github.com/ruby/debug/issues/336#issuecomment-968349278


jejacks0n:
Hey! Thanks for this.

